PR #24, opened 2019-11-29
Dispatch repository events to Integration Tests repository

manusa:
A workflow that will trigger Integration tests in a remote repository (See: https://github.com/jshiftio/jkube-integration-tests/blob/master/.github/workflows/integration-tests.yml).

There is an **alternative** approach to perform the opposite of the current workflow so that integration tests run in this repository. This would imply creating a workflow in JKube that would clone the integration tests repository so that tests could be run in the workflow's context.
GitHub checks would also be enabled in this scenario.
**Please comment if you think this would be a better solution**

Relates to: https://github.com/jkubeio/jkube/issues/21

### DIFF
--- a/.github/workflows/repository-dispatcher.yml
+++ b/.github/workflows/repository-dispatcher.yml
@@ -1,0 +1,29 @@
+#
+# Copyright (c) 2019 Red Hat, Inc.
+# This program and the accompanying materials are made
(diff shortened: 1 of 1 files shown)


manusa:
After discussion in team chat group, I'll decline this PR and reimplement using the alternative approach mentioned in the previous comment (pull strategy)

PR #26, opened 2019-12-12
fix: dependency version convergence error for direct dependency

manusa:
We have a direct dependency to `org.yaml:snakeyaml:1.25` and `com.fasterxml.jackson.dataformat:jackson-dataformat-yaml` depends on `1.23`.
This change excludes the latter.

### DIFF
--- a/jkube-kit/common/pom.xml
+++ b/jkube-kit/common/pom.xml
@@ -47,6 +47,11 @@
     </dependency>
 
     <dependency>
+      <groupId>org.apache.commons</groupId>
+      <artifactId>commons-compress</artifactId>
+    </dependency>
+
+    <dependency>
       <groupId>com.google.code.gson</groupId>
       <artifactId>gson</artifactId>
     </dependency>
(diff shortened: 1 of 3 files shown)


manusa:
I fixed 2 more convergence errors.
In order to fix one of them I had to upgrade this dependency: `org.bouncycastle:bcpkix-jdk15on:1.61`
We'll need to file a new CQ for this one.

rohanKanojia:
Hmm, I see. Can we hold this a bit until we cut our first release? What do you think? Or do you think it's very critical?

I'm not sure whether CI failure is related to this change :confused:

manusa:
Not critical at all. I'm doing this to remove the number of exclusions for the quarkus/kubernetes-extension PoC.
In the future, we should also probably add the maven enforcer plugin. We have many dependencies right now, and many of their transitive dependencies have convergence errors (which may lead to non-deterministic bugs).

Regarding CI failure, has nothing to do with the project. There's probably an error in some of Microsoft's APT repos (https://github.community/t5/GitHub-Actions/ubuntu-latest-Apt-repository-list-issues/td-p/41122). Minkube setup action is not able to complete.

manusa:
Outdated by #39

Issue #30, opened 2019-12-18
Decouple JKube-Kit from Maven


manusa:
## Description
In order to be able to use JKube as an external library, JKube kit should be completely decoupled from any Maven dependencies

## Tasks
- [x] Remove all non-needed dependencies from `jkube-kit-common` and replicate them in the projects where needed -> #44
- [x] Move maven specific utils from `jkube-kit-common` to a separate "Maven only" project -> #44 
- [x] Remove maven dependencies from jkube-kit-config-image -> #45 
- [x] Move resource service implementation out of config model-> #46
  - Create module `resource/service` "Jkube Kit :: Resource :: Service"
  - Move package `org.eclipse.jkube.kit.config.service` in `config/resource` to the new module
- [x] Remove maven dependencies from jkube-kit-config-resource -> #46 
- [x] Remove maven dependencies from jkube-kit-build-api and jkube-kit-build-service-docker -> #48 
- [x] Remove maven dependencies from jkube-kit-build-maven -> #49
- ...

## Blocked by
- #29 #39

rohanKanojia:
@manusa : oh, Did we miss something?

PR #44, opened 2020-01-09
refactor: jkube-kit-common decoupled from Maven

manusa:
#30 First task: decouple jkube-kit-common from Maven


jkube-kit/common/src/test/java/org/eclipse/jkube/kit/common/util/EnvUtilTest.java
```diff
@@ -0,0 +1,52 @@
+package org.eclipse.jkube.kit.common.util;
```
rohanKanojia: license header missing

jkube-kit/common-maven/src/main/java/org/eclipse/jkube/kit/common/util/MavenUtil.java
```diff
@@ -274,6 +277,27 @@ public static String getVersion(String groupId, String artifactId) throws IOExce
         return version;
     }
 
+    /**
+     * Return all properties in Maven project, merged with all System properties (-D flags sent to Maven).
+     * <p>
```
rohanKanojia: looks like this came through autoformat

manusa: This was already here:
https://github.com/eclipse/jkube/blob/08a546bb0dbb8bf6cf2949a8f549193b144b848e/jkube-kit/common/src/main/java/org/eclipse/jkube/kit/common/util/EnvUtil.java#L228

I don't know if we have any javadoc guidelines, I usually prefer the way defined in Spring code style guidelines (but doesn't make any difference):
```
Return all properties...

<p> System proper
```

Most of our code is using this style, so I would let it as it is.

PR #45, opened 2020-01-13
refactor: jkube-kit-config-image decoupled from Maven

manusa:
#30 Second task: decouple jkube-kit-config-image from Maven


jkube-kit/config/image/src/main/java/org/eclipse/jkube/kit/config/image/build/BuildConfiguration.java
```diff
@@ -5,7 +5,7 @@
  * which is available at:
  *
  *     https://www.eclipse.org/legal/epl-2.0/
- *
+ *AssemblyConfiguration
```
rohanKanojia: typo?

jkube-kit/config/image/src/main/java/org/eclipse/jkube/kit/config/image/build/BuildConfiguration.java
```diff
@@ -283,10 +280,9 @@ public boolean optimise() {
         return optimise != null ? optimise : false;
     }
 
-
     @Deprecated
     public String getCommand() {
```
rohanKanojia: Shall we remove the method also?

jkube-kit/generator/java-exec/src/test/java/org/eclipse/jkube/generator/javaexec/JavaExecGeneratorMainClassDeterminationTest.java
```diff
@@ -14,27 +14,20 @@
...
 
 import java.io.File;
-import java.io.IOException;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.List;
-import java.util.jar.Attributes;
 
-import com.fasterxml.jackson.databind.util.ClassUtil;
+import mockit.*;
```
rohanKanojia: Could you please void wildcard imports?

jkube-kit/config/image/src/main/java/org/eclipse/jkube/kit/config/image/build/BuildConfiguration.java
manusa: :open_mouth: Nice catch!

jkube-kit/config/image/src/main/java/org/eclipse/jkube/kit/config/image/build/BuildConfiguration.java
manusa: In JKube it's only used in `DockerAssemblyManager`.
I didn't want to remove it because it's part of the public API, but anyway, I don't think anyone is using JKube-kit besides ourselves, so I'll remove.

Issue #58, opened 2020-02-06
Build failure when building a docker image with image configuration.


rohanKanojia:
I tried building a docker image with this configuration:
```
      <plugin>
        <groupId>org.eclipse.jkube</groupId>
        <artifactId>k8s-maven-plugin</artifactId>
        <version>0.1.1-SNAPSHOT</version>
        <configuration>
          <images>
            <image>
              <!-- Add your dockerhub user name here -->
              <name>rohankanojia/helloworld-java:${project.version}</name>
              <alias>hello-world</alias>
              <build>
                <from>openjdk:latest</from>
                <assembly>
                  <descriptorRef>artifact</descriptorRef>
                </assembly>
                <cmd>java -jar maven/${project.name}-${project.version}.jar</cmd>
              </build>
              <run>
                <wait>
                  <log>Hello World!</log>
                </wait>
              </run>
            </image>
          </images>
        </configuration>
      </plugin>
```

However, `k8s:build` goal fails giving this error:
```
~/work/repos/jkube-imageconfig-sample : $ mvn k8s:build -X
Apache Maven 3.6.3 (cecedd343002696d0abb50b32b541b8a6ba2883f)
Maven home: /opt/apache-maven
Java version: 1.8.0_231, vendor: Oracle Corporation, runtime: /usr/java/jdk1.8.0_231-amd64/jre
Default locale: en_US, platform encoding: UTF-8
OS name: "linux", version: "5.4.13-201.fc31.x86_64", arch: "amd64", family: "unix"
[DEBUG] Created new class realm maven.api
[DEBUG] Importing foreign packages into class realm maven.api
[DEBUG]   Imported: javax.annotation.* < plexus.core
[DEBUG]   Imported: javax.annotation.security.* < plexus.core
[DEBUG]   Imported: javax.enterprise.inject.* < plexus.core
[DEBUG]   Imported: javax.enterprise.util.* < plexus.core
[DEBUG]   Imported: javax.inject.* < plexus.core
[DEBUG]   Imported: org.apache.maven.* < plexus.core
[DEBUG]   Imported: org.apache.maven.artifact < plexus.core
[DEBUG]   Imported: org.apache.maven.classrealm < plexus.core
[DEBUG]   Imported: org.apache.maven.cli < plexus.core
[DEBUG]   Imported: org.apache.maven.configuration < plexus.core
[DEBUG]   Imported: org.apache.maven.exception < plexus.core
[...]

PR #62, opened 2020-02-12
fix: Problem with plexus detecting target value type for bean injection

manusa:
After the initial Maven decoupling commits, a bug has appeared specific to Plexus injection of generic typed bean values.

This fix allows to inject `<configuration><images><image><build><assembly> ` entries again.


manusa:
Related to #58

PR #103, opened 2020-03-12
refactor: openshift-maven-plugin inherits from kubernetes-maven-plugin

### LINKED ISSUE #53: Rename Plugins' artifactIds to kubernetes-maven-plugins and openshift-maven-plugins
Right now we have plugin `artifactIds` set to `k8s-maven-plugin` and `oc-maven-plugin`. We should rename it to `kubernetes-maven-plugin` since it would make it easier to get searched on maven central. But we should also preserve current plugin prefixes (`k8s` and `oc`). It's possible to do it as mentioned on this StackOverflow [post](https://stackoverflow.com/questions/2736373/how-to-rename-goals-in-maven).

manusa:
Closes #53


openshift-maven-plugin/plugin/src/main/java/org/eclipse/jkube/maven/plugin/mojo/Openshift.java
```diff
@@ -0,0 +1,9 @@
+package org.eclipse.jkube.maven.plugin.mojo;
```
rohanKanojia: license header missing

PR #105, opened 2020-03-16
Fix NullPointerException in DockerAssemblyManager

rohanKanojia:
I noticed this while running hello-world sample when some
user pinged us on gitter that it's broken. The user seemed to be
using an older version but we're facing NullPointerException on v0.2.0


manusa:
Did you change quickstarts pom versions by running `./scripts/quickstarts.sh version`. If not, please just run it to make sure all versions are changed consistently.

rohanKanojia:
I ran the script, looks like everything was modified accordingly.

Issue #113, opened 2020-03-19
helm : enhance chart generation


survivant:
I would like to have the option to choose where I want to output charts to be generated.  I'll like prefer to have it in ./charts folder instead of target...  like that I could add it to my source code.

something like : 

```
<plugin>
        <groupId>org.eclipse.jkube</groupId>
        <artifactId>k8s-maven-plugin</artifactId>
        <version>0.2.0</version>
         <configuration>
              <outputDir>./charts</outputDir>
          </configuration>
      </plugin>

for now I'm using this 

(your helm mojo is broken for now, that's why I'm using fabric8)
mvn fabric8:helm -D=fabric8.helm.outputDir=./charts
```

the file Chart.yaml could be enhanced like that (it contains more useful informations and display nice in chartmuseum UI too : 

```
apiVersion: v1
appVersion: "1.0"
description: A Helm TEST chart for Kubernetes
name: test-k8s-helm
version: 0.1.1
maintainers: me@cherche.pas
home: http://url-du-projet
icon: http://littleicon.png
sources:
  - githubpath
```

all that information could be in configuration in the pom.xml

if the folder is present.. just override the files without deleted the content.

rohanKanojia:
Thanks for your bug report. @manusa is working on fixing helm support. It would be fixed before the next release

rohanKanojia:
related to #102

manusa:
Implementation fix includes options to specify both `outputDir` where all chart files will be generated, and `tarballOutputDir` where the tar archive will be generated (independent from `outputDir`)

Both options can be setup within pom.xml configuration or overridden with properties (e.g. `jkube.helm.outputDir`)

Regarding `outputDir` behavior, generated charts will be output by default in `target/$chartName/`, so for each helmType (one of `kubernetes` or `openshift`) charts and files will be output in `$outputDir/$helmType`. As said the value of `outputDir` can be overridden, but every generated file will always be nested under `$outputDir/kubernetes` or `$outputDir/openshift`.

Regarding `tarballOutputDir` will default to `target`.

So implementation differs a little with that of FMP, hope this is OK for you. Please if you find something wrong with this or new behavior is of great inconvenience post a reply with your thoughts.

manusa:
> if the folder is present.. just override the files without deleted the content.

Seems fair for this to be the default behavior, as mvn clean should take care of cleaning. I'll modify current behavior

manusa:
Hi @survivant 

[...]

Issue #102, opened 2020-03-12
HelmMojo no longer works


manusa:
## Description

HelmMojo expects an additional Plexus dependency

`java.lang.NoClassDefFoundError: org/codehaus/plexus/components/io/resources/PlexusIoResourceCollection`

If possible, refactor HelmMojo to be independent from Maven. Later on, implementation should be moved to JKube-Kit so it's reusable by other plugins, extensions, etc.

- [X] Port missing changes from fabric8io/fabric8-maven-plugin#1642
- [X] Remove Maven dependencies
- [x] Move to JKube Kit > Helm
- [x] Helm generation doesn't clean target directories
- [x] Inherit Helm Mojo in openshift-maven-plugin
- [x] Create Quickstart
- [x] Add Unit Tests
- [x] Add integration tests (https://github.com/jkubeio/jkube-integration-tests)
- [x] Update documentation

## Relates to:
- #30 
- #83

PR #116, opened 2020-03-23
Refactor HelmMojo

manusa:
## Description
- Moved all Helm logic to JKube Kit so it's accessible programmatically (Can be used from Gradle Plugins, Maven Plugins, Extensions, etc.
- Fixed current bugs
- Added extra-properties to HelmMojo
- Ensured Maven HelmMojo configuration precedence as: Maven property > pom.xml configuration > opinionated default (so if there is a pom configuration, it will be used instead of a default, if there is a property too the property will be used instead)
- Added Unit tests
- Added complete integration test to HelmService > This test demonstrates all HelmService features (template generation, variable substitution, tar archiving, etc.)
- Added Quickstart

## Relates to:
- #102 
- #113


rohanKanojia:
Thanks a lot for this :heart: . Could you also update our helm goal documentation with new changes?

manusa:
> Thanks a lot for this  . Could you also update our helm goal documentation with new changes?

Still work in progress. I still have to add some integration tests in the other repo (https://github.com/jkubeio/jkube-integration-tests), but I need this one to be merged so I can start working on that.

Anyway, I'll add an additional task in #102

quickstarts/maven/spring-boot-with-yaml/pom.xml
```diff
@@ -0,0 +1,169 @@
...
+              </resources>
+              <generator>
+                <config>
+                  <spring-boot>
+                    <color>always</color>
+                  </spring-boot>
+                </config>
+              </generator>
+              <enricher>
+              </enricher>
```
rohanKanojia: What's the purpose of this empty `<enricher>` block?

quickstarts/maven/spring-boot-with-yaml/pom.xml
```diff
@@ -0,0 +1,169 @@
...
+<!--                <sources>-->
+<!--                  <source>https://different.source/yaml.yml</source>-->
+<!--                </sources>-->
+<!--                <maintainers>-->
+<!--                  <maintainer>-->
+<!--                    <name>John</name>-->
+<!--                    <email>john.doe@example.com</email>-->
+<!--                  </maintainer>-->
+<!--                </maintainers>-->
+<!--              </helm>-->
```
rohanKanojia: Is it okay to leave it commented here like this? I was thinking if we could have a dedicated sample for helm. Maybe we can rename this sample to `spring-boot-helm`... WDYT?

manusa: The name was inherited from FMP, I can rename to helm (it's clearer to me than "-yaml"

Regarding the commented-out section, this is a Quickstart. The purpose is to showcase how the plugin behaves given certain configurations, we can either create more granular quickstarts, add more profiles, or leave commented-out sections such as this one.

I opted for this approach because creating too many Quickstarts increase the maintenance complexity and also may pollute the code-base if there is no sort of criteria when defining the examples.
Using maven profiles is also a way to achieve this, but the pom will be more difficult to read and get messy.
Using commented out sections has the advantage that we can actually show what configurations are supported by the Mojo with examples and the user can un-comment each configuration to see how it behaves. This is not production code but sample code it should be verbose and documented.

The  `<enricher></enricher>` must be removed.

quickstarts/maven/spring-boot-with-yaml/pom.xml
manusa: It has to be removed

PR #124, opened 2020-03-26
Fix #75: Configuration mismatch between docs and code

### LINKED ISSUE #75: Configuration mismatch between docs and code
The XML configuration in the docs (even in Fabric8 ones), does not seem to match what the code does. This was a source of confusion when I started using these plugins initially (until I dug through the source code).

For example, the [XML Configuration](https://htmlpreview.github.io/?https://github.com/eclipse/jkube/blob/master/kubernetes-maven-plugin/doc/index.html#xml-configuration) outlines what `<resources>` item looks like but this does not match the [`ResourceConfig`](https://github.com/eclipse/jkube/blob/master/jkube-kit/config/resource/src/main/java/org/eclipse/jkube/kit/config/resource/ResourceConfig.java) class.

If I provide say the `<deployment>` child as outlined in that section, I get the following error:

```
[ERROR] Failed to execute goal org.eclipse.jkube:k8s-maven-plugin:0.1.1:resource (default-cli) on project pdfg: Unable to parse configuration of mojo org.eclipse.jkube:k8s-maven-plugin:0.1.1:resource for parameter deployment: Cannot find 'deployment' in class org.eclipse.jkube.kit.config.resource.ResourceConfig -> [Help 1]
```

But when I do follow the class structure, everything goes as expected.

Am I missing something here?

rohanKanojia:
Fix #75 

Updated docs as per existing codebase

### DIFF
--- a/kubernetes-maven-plugin/doc/src/main/asciidoc/inc/_introduction.adoc
+++ b/kubernetes-maven-plugin/doc/src/main/asciidoc/inc/_introduction.adoc
@@ -144,7 +144,7 @@ mvn k8s:resource k8s:deploy
 
 By default a _Service_ and a _Deployment_ object pointing to the created Docker image is created. When running in OpenShift mode, a _Service_ and _DeploymentConfig_ which refers the _ImageStream_ created with `k8s:build` will be installed.
 
-Of course you can bind all those jkube.goals to execution phases as well, so that they are called along with standard lifecycle goals like `install`. For example, to bind the building of the Kubernetes resource files and the Docker images, add the following goals to the execution of the f-m-p:
+Of course you can bind all those jkube.goals to execution phases as well, so that they are called along with standard lifecycle goals like `install`. For example, to bind the building of the Kubernetes resource files and the Docker images, add the following goals to the execution of the k-m-p:
 
 .Example for lifecycle bindings
 [source, xml, indent=0]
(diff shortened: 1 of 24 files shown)


kubernetes-maven-plugin/doc/src/main/asciidoc/inc/_introduction.adoc
```diff
@@ -237,31 +238,10 @@ A working example can be found in the  https://github.com/jkubeio/kubernetes-mav
...
-          <type>emptyDir</type>
-        </volume>
-      </volumes>
-    </deployment>
-
-    <services> <!--8-->
+    <replicas>2</replicas> <!--4-->
+    <controllerName>${project.artifactId}</controllerName> <!--5-->
+
+    <services> <!--6-->
```
manusa: What happened to these configurations, why have we lost support for this?

manusa:
Table organization is inconsistent all around. We need to unify this:
I see examples of:
image
where configuration property is first.

And:
image
image
where configuration property is last.

Tables with configuration information should be consistent across the docs: same number of columns, in the some order and with analogue content.

kubernetes-maven-plugin/doc/src/main/asciidoc/inc/_introduction.adoc
rohanKanojia: I looked for old plugin versions but I couldn't find any config for containers or deployments.

openshift-maven-plugin/doc/src/main/asciidoc/inc/_introduction.adoc
```diff
@@ -144,7 +144,7 @@ mvn oc:resource oc:deploy
 
 By default a _Service_ and a _Deployment_ object pointing to the created Docker image is created. When running in OpenShift mode, a _Service_ and _DeploymentConfig_ which refers the _ImageStream_ created with `oc:build` will be installed.
 
-Of course you can bind all those jkube.goals to execution phases as well, so that they are called along with standard lifecycle goals like `install`. For example, to bind the building of the Kubernetes resource files and the Docker images, add the following goals to the execution of the f-m-p:
+Of course you can bind all those jkube.goals to execution phases as well, so that they are called along with standard lifecycle goals like `install`. For example, to bind the building of the Kubernetes resource files and the Docker images, add the following goals to the execution of the kmp:
```
manusa: We're in OpenShift documentation, see line 5.
Same applies here, `oc-m-p` `o-m-p` or any other permutation should be replaced by `omp` (for consistency with the `kmp` chosen in Kubernetes Maven Plugin docs).

Issue #131, opened 2020-03-30
Windows: jar artifact generated in wrong directory


manusa:
Relates to #113 

> mvn k8s:build generate a artifcat in the root folder : targetquarkus-k8s-hello-1.0.0-SNAPSHOT.jar there is a "/" missing. should be : target/quarkus-k8s-hello-1.0.0-SNAPSHOT.jar

> k8s:build generates the jar file in the wrong folder : missing / to get target/quarkus-k8s-hello-1.0.0-SNAPSHOT.jar

PR #140, opened 2020-04-02
fix: dependency version convergence error

manusa:
Relates to: #131


manusa:
> Umm, How is this fixing #131 ?

:o the change got lost after rebase, It was affected by the line that was changed yesterday:
https://github.com/rohanKanojia/jkube/blob/77dc2a888ae0d786d5d5b1ad5463cd8e59ea7b83/jkube-kit/build/service/docker/src/main/java/org/eclipse/jkube/kit/build/core/assembly/DockerAssemblyManager.java#L404

PR #145, opened 2020-04-08
added documentation for jkube-controller enricher

zebu:
Added documentation to cover the defaults for the jkube-controller enricher.

I followed the jkube-service as an example but I do see some repetition in this line: 
> This can be either directly configured with fragments or with the XML configuration
In the future, it may be better to have a section that describes the general behavior of configuration options and the precedence of operations.

I supplied links to the K8s controller documentation but I could not see how to link to the list on controllers which I'm not happy about. 

https://stackoverflow.com/questions/60918319/

### DIFF
--- a/kubernetes-maven-plugin/doc/src/main/asciidoc/inc/_enricher.adoc
+++ b/kubernetes-maven-plugin/doc/src/main/asciidoc/inc/_enricher.adoc
@@ -50,7 +50,7 @@ kubernetes-maven-plugin comes with a set of enrichers which are enabled by defau
 | Add Maven SCM information as annotations to the kubernetes/openshift resources
 
 | <<jkube-controller>>
-| Create default controller (replication controller, replica set or deployment) if missing.
+| Create default controller (replication controller, replica set or deployment https://kubernetes.io/docs/concepts/workloads/controllers/[Kubernetes doc]) if missing.
 
 | <<jkube-dependency>>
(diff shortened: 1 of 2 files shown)


manusa:
Hi @zebu,

Thanks for checking out the project and contributing. In order to be able to merge your PR, you need to sign-off your commit with an e-mail account belonging to a user who has signed the Eclipse Contributor Agreement.

More details: https://www.eclipse.org/jkube/contributing#legal

zebu:
Thanks, it's signed. The issue is the commit has the wrong email so it's not matching the ECA. It's late. I'll sort it out tomorrow.

rohanKanojia:
@zebu : ah, why did you close your pull request? Is something wrong?

zebu:
I think the request automatically closed when I did a hard reset and pushed? I'm not really sure. It took me a while to figure out how to fix the commiter email in github. Now it looks like I have a different eca issue to fix.

zebu:
I'm not sure how to fix this and I'm making a mess of it. I have 3 commits now. Two are signed, one is not. Of the two signed, only one doesn't have an 'X'.  I'm not a strong git user. What's the best way to correct? Should I start over, hard rest back to master, and commit my change with the sign-off?

rohanKanojia:
I think you can do a `git rebase -i HEAD~3` to combine all commits into one. And then `git commit --amend --author="Your name<youremail@emailprovider.com>"`

Issue #150, opened 2020-04-14
no docker container created for hello-world


codezork:
If you try to run the hello-world sample with `mvn clean install k8s:build k8s:resource k8s:apply` the pod's status will show `helloworld-5f9d89f5c4-npjz9   0/1     ImagePullBackOff   0          17h` as there is no docker image in the repository. The log says: `Error from server (BadRequest): container "hello-world" in pod "helloworld-5f9d89f5c4-npjz9" is waiting to start: trying and failing to pull image`.
[pom.txt](https://github.com/eclipse/jkube/files/4473752/pom.txt)

manusa:
If you are running with Minikube, remember to `eval $(minikube docker-env)` so that the Docker local image registry is shared with the cluster.

If running a regular K8s cluster, you need to push the image to repository so that the cluster can actually download the image.

codezork:
Yep. This did the trick. Thank you. Why not mentioning at Readme.md? Fantastic job BTW, fabric8 and jkube!

rohanKanojia:
@codezork : umm, Would you like to create a quick PR to fix our doc :wink: ?

codezork:
LOL - why not? What about ISTIO? Do you have plans?

rohanKanojia:
Umm, not so sure about Istio. I think we should definitely consider it. There was an issue filed on FMP regarding this https://github.com/fabric8io/fabric8-maven-plugin/issues/1164 . Maybe we should consider it after releasing JKube 1.0.0

codezork:
hmm. Didn't sound like a show-stopper. Would be nice for a canary-release option.

codezork:
PR done

PR #151, opened 2020-04-15
Update README.md

codezork:
Added 'Troubleshooting' sharing minikube's docker env

### DIFF
--- a/kubernetes-maven-plugin/doc/src/main/asciidoc/inc/_enricher.adoc
+++ b/kubernetes-maven-plugin/doc/src/main/asciidoc/inc/_enricher.adoc
@@ -50,7 +50,7 @@ kubernetes-maven-plugin comes with a set of enrichers which are enabled by defau
 | Add Maven SCM information as annotations to the kubernetes/openshift resources
 
 | <<jkube-controller>>
-| Create default controller (replication controller, replica set or deployment) if missing.
+| Create default controller (replication controller, replica set or deployment https://kubernetes.io/docs/concepts/workloads/controllers/[Kubernetes doc]) if missing.
 
 | <<jkube-dependency>>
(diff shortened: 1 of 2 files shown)


rohanKanojia:
Thanks for your quick PR :+1: . In order to be able to merge your PR, you need to sign-off your commit with an e-mail account belonging to a user who has signed the Eclipse Contributor Agreement.

More details: https://www.eclipse.org/jkube/contributing#legal

manusa:
Relates to #150

codezork:
can I also set signoff from the ui? I didn't check out for the readme change.

Shouldn't we change the pom.xml to version "1.0.0-alpha-1" for at least in helloworld?

rohanKanojia:
I think you would need to create an eclipse account to be able to sign ECA. Actually all of our quickstarts are at SNAPSHOT  version right now. @manusa What do you think about setting quickstarts version to latest release rather than SNAPSHOTS?

manusa:
> can I also set signoff from the ui? I didn't check out for the readme change.

Sign off is just an extra-line in the commit message with the format "Signed-off-by: Your Name <email@example.com>

> Shouldn't we change the pom.xml to version "1.0.0-alpha-1" for at least in helloworld?

Versions in Quickstarts are modified using a script so that they all have the same version, usually this is done upon release and tag.

`./scripts/quickstarts.sh version`

The project is currently under very active development, so we committed an extra change to point quickstarts back to SNAPSHOT.

I really don't know if there are many users now, depending on that it may be helpful to change quickstarts version back to current release.

rohanKanojia:
@codezork : ah, not sure why PR got closed on update :broken_heart: . Could you please create a separate PR with the corrected email which is a registered Eclipse account?

I admit it's a bit frustrating to deal with ECA :-(

Issue #90, opened 2020-03-04
Update JKube Kit diagrams


manusa:
## Description

After the completion of #30, JKube Kit model diagrams should be updated to represent the current status of the project.

https://github.com/eclipse/jkube/tree/master/jkube-kit/doc

PR #161, opened 2020-04-23
doc: #90 Updated module overview diagram to current status

manusa:
Closes: #90 
module_overview

### DIFF
--- a/jkube-kit/doc/src/main/asciidoc/inc/diagrams/module-overview.puml
+++ b/jkube-kit/doc/src/main/asciidoc/inc/diagrams/module-overview.puml
@@ -45,77 +45,83 @@ package jkube.kit {
     package kit.config {
         [config-image] --> [common]
         [config-resource] --> [common]
+        [config-service] --> [config-resource]
+        [config-service] --> [build-service-docker]
     }
     package kit.build {
         [build-api] --> [config-image]
-        [build-maven] --> [build-api]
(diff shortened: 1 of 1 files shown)


rohanKanojia:
Could you please update README image link too?

manusa:
> Could you please update README image link too?

I don't think this model replaces the one on the readme. The intention of this one is to show current status. Which differs  a lot from what the one in the readme shows.

We should work with this one to establish next steps in the project (especially profiles).

rohanKanojia:
For showing current status, it needs to be somewhere. Maybe inside jkube-kit's README. I'm not sure if someone interested in project would look into target folders for the diagram

manusa:
I think what you mention belongs more to 
https://github.com/eclipse/jkube/tree/master/jkube-kit/doc (current root folder)

Or even some kind of formal Roadmap (which we don't yet have).

I would rather create a new issue to update/refactor current JKube Kit docs or to create a Roadmap, and maybe also embed them in the website (just like the maven plugins). But wouldn't do it in scope of this one.

I initially created the diagram to design a **phased** decoupling from Maven (which we didn't do in the end as it was performed in a single run).

I think this diagram is still important (that's why it's being maintained) because we can compare it to Roland's original design and see what we're missing (e.g. profiles). We can use this internally to decide on next steps and prioritize upcoming tasks.

rohanKanojia:
Umm, okay. Let me merge this then

Issue #183, opened 2020-05-07
[EPIC] Change base images


manusa:
## Description
As we move away from fabric8 organization, it's good not to depend on fabric8 maintained [base images](https://github.com/fabric8io-images).

Since [Red Hat Universal Base Image (UBI)](https://jmtd.net/log/ubi-openjdk/) was recently released, we can switch _all_ of [our base images](https://github.com/eclipse/jkube/blob/9e3d67371e8c12bc8aa6288acacc83d1f796ea96/jkube-kit/parent/pom.xml#L78) with those provided by UBI.

- https://access.redhat.com/containers/?tab=tech-details&platform=openshift&get-method=unauthenticated#/registry.access.redhat.com/ubi8/openjdk-8

- https://access.redhat.com/containers/?tab=tech-details&platform=openshift&get-method=unauthenticated#/registry.access.redhat.com/ubi8/openjdk-11

- http://post-office.corp.redhat.com/archives/thecore/2020-May/msg00084.html

These are the [base images](https://github.com/eclipse/che-dockerfiles) Eclipse Che is using for [Devfiles](https://github.com/eclipse/che-devfile-registry).

These are the images [quarkus](https://github.com/quarkusio/quarkus-images) uses. For Java-Exec we can implement our own with a similar approach.

## Current Base Images
| Image type | Usage (Generator) | Current Image | Proposed Image |
|-|-|-|-|
| java.upstream.s2i | java-exec,spring-boot,wildfly-swarm | fabric8/s2i-java | quay.io/jkube/jkube-java-binary-s2i #205 |
| java.upstream.docker | java-exec,spring-boot,wildfly-swarm | fabric8/java-centos-openjdk8-jdk | quay.io/jkube/jkube-java-binary-s2i #205 |
| java.redhat.s2i | java-exec,spring-boot,wildfly-swarm | jboss-fuse-6/fis-java-openshift | No longer applicable #220 |
| java.redhat.docker | java-exec,spring-boot,wildfly-swarm | jboss-fuse-6/fis-java-openshift | No longer applicable #220 |
| karaf.upstream.docker | karaf | fabric8/s2i-karaf | quay.io/jkube/jkube-karaf-binary-s2i #214 |
| karaf.upstream.s2i | karaf | fabric8/s2i-karaf | quay.io/jkube/jkube-karaf-binary-s2i #214 |
| karaf.redhat.docker | karaf | jboss-fuse-6/fis-karaf-openshift | No longer applicable #220 |
[...]

PR #203, opened 2020-05-15
spike: JavaExecGenerator uses ubi8/openjdk-11 always

manusa:
This is PoC PR intended for review of options regarding base image replacements (#183)

What this PR accomplishes is the use of `registry.access.redhat.com/ubi8/openjdk-11` image for all build strategies.
The main advantage for this is the reduced complexity in maintenance, as configuration should only deal with one image type.

We're also replacing 2 different fabric8 maintained images for a single community-driven image.

In order to be able to use the s2i builder image both as a builder image and as a runtime image is by providing the packaged artifacts in the specific directory for s2i assembly (`/tmp/src/deployments`). And by adding the COPY statement from deployments to `/deployments` in the generated Dockerfile.


rohanKanojia:
@manusa : Is this ready for review?

manusa:
This is closed by #205.

It was a PoC showcasing how to use unmodified community-based upstream images.

#205 approach is preferred because 100 MiB are stripped off of the upstream image and not much maintenance is required since it's based in community-based CEKit modules.

Issue #188, opened 2020-05-07
Create Apache Karaf Maven Quick start


manusa:
## Description

Create a Quick start that enables us to test/develop `KarafGenerator`.

The quick start should also include some documentation describing how to run, configure, etc. This will allow users to see what's the recommended approach when integration Eclipse JKube with Apache Karaf related projects.

**c.c.**: @anandrkskd

rohanKanojia:
There is one karaf based sample using FMP: https://github.com/fabric8-quickstarts/karaf-camel-log

manusa:
TIL: https://github.com/fabric8-quickstarts

manusa:
https://github.com/eclipse/jkube/tree/master/quickstarts/maven/karaf-camel-log

Issue #94, opened 2020-03-05
Properly define + document JKubeProjectAssembly behavior


manusa:
## Description

As of now, JKubeProjectAssembly behavior is not aligned with [Maven's Assemby Plugin](https://maven.apache.org/plugins/maven-assembly-plugin/assembly.html) behavior.

Behavior for filesets includes, excludes, filtered, etc. is opinionated and doesn't perform as described in https://maven.apache.org/plugins/maven-assembly-plugin/assembly.html#class_fileSet.
There are some fields missing too.

Same goes for [file](https://maven.apache.org/plugins/maven-assembly-plugin/assembly.html#class_file), [repository](https://maven.apache.org/plugins/maven-assembly-plugin/assembly.html#class_repository), etc.

We should either properly document our API behavior or completely implement the behavior mimicking the one provided by Apache Maven Plugin.

## Relates to:
- #87 
- https://books.sonatype.com/mvnref-book/reference/assemblies-sect-basics.html
- https://books.sonatype.com/mvnref-book/reference/assemblies-sect-controlling-contents.html

manusa:
## [WIP]Current status:

### [fileSet](https://maven.apache.org/plugins/maven-assembly-plugin/assembly.html#class_fileSet) - AssemblyFileSet
- Discarded properties:
   - ~useDefaultExcludes~
   - ~lineEnding~
- Supported properties
   - outputDirectory
   - includes
   - excludes
   - fileMode
   - directoryMode
   - directory
- Additional properties
   - outputDirectory
- Pending
   - filtered
   - nonFilteredFileExtensions

PR #214, opened 2020-05-28
fix: KarafGenerator update

manusa:
- Created Karaf Quickstart
- Fixed problems with FileSet assembly config to include complete directories
- Upgraded images to use quay.io/jkube/jkube-karaf-binary-s2i

**n.b.** `CompleteK8sITCase` will fail due to an invalid assertion. In the past filesets entries were copied (in all cases) to the target directory under a subdirectory named like the source parent directory, which is wrong.)
The failing line should be replaced
- From:
```
assertThat(imageFiles, hasItem("/deployments/jkube-includes/will-be-included-if-no-assemblies-defined.txt"));
```
- To:
```
    assertThat(imageFiles, hasItem("/deployments/will-be-included-if-no-assemblies-defined.txt"));
```

I'll update integration-tests once this PR gets merged.

Relates to: #183 #188 #94


jkube-kit/common/src/test/java/org/eclipse/jkube/kit/common/archive/AssemblyFileSetUtilsProcessAssemblyFileSetTest.java
```diff
@@ -0,0 +1,343 @@
...
+        .build();
+    // When
+    final Map<File, String> permissions = AssemblyFileSetUtils.processAssemblyFileSet(baseDirectory, outputDirectory, afs, ac);
+    // Then
+    assertThat(permissions.entrySet(), hasSize(4));
+    final File deployments = new File(outputDirectory, "deployments");
+    assertThat(deployments.exists(), equalTo(true));
+    assertThat(deployments.listFiles(), arrayWithSize(6));
+    assertThat(deployments.list(), arrayContainingInAnyOrder("one", "two", "three", "1.txt", "3.other", "37"));
+    final File outputSourceDir = Objects.requireNonNull(deployments.listFiles())[0];
```
rohanKanojia, line 213: `outputSourceDir` seems unused. Umm, maybe you forgot to assertions for it? If not, could you please remove it?

manusa: Yep, some descriptions are wrong too. I changed the tests when I figured out some scenarios were wrongly defined.

PR #216, opened 2020-06-01
Fix #195: Added MigrateMojo for migrating projects from FMP

### LINKED ISSUE #195: Plugin MigrateMojo to automate migration from FMP to JKube
## Description

In order to increase the adoption of the plugin we can create a `MigrateMojo` that will automatically update any `pom.xml` so that it uses Eclipse JKube instead of Fabric8 Maven Plugin.

Basically, any time the command `mvn org.eclipse.jkube:kubernetes-maven-plugin:migrate` is run in a Maven project root directory, the `pom.xml` file is parsed and appropriate substitutions are made.

Ideally, MigrateMojo should be "smart" enough to see if `fabric8-maven-plugin` should be replaced by `kubernetes-maven-plugin` or `openshift-maven-plugin`.

If the previous requirement is not possible or very difficult to implement, we can have an OpenshiftMigrateMojo that extends the basic MigrateMojo and overrides `getArtifactId` method to return `openshift-maven-plugin` instead of the super implementation `kubernetes-maven-plugin`.

The plugin should try to consider all configurations, flags and properties and rename them to match the new ones.

## See also
- There was a [MigrateMojo](https://github.com/fabric8io/fabric8/blob/874aeddd493c16133366d3ec824685d07479dbdf/fabric8-maven-plugin/src/main/java/io/fabric8/maven/MigrateMojo.java) in FMP v3, maybe there's something useful there (or maybe not)

rohanKanojia:
Fix #195

### DIFF
--- a/jkube-kit/common/src/main/java/org/eclipse/jkube/kit/common/service/MigrateService.java
+++ b/jkube-kit/common/src/main/java/org/eclipse/jkube/kit/common/service/MigrateService.java
@@ -1,0 +1,152 @@
+/**
+ * Copyright (c) 2019 Red Hat, Inc.
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at:
+ *
+ *     https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *   Red Hat, Inc. - initial API and implementation
+ */
+package org.eclipse.jkube.kit.common.service;
+
+import org.eclipse.jkube.kit.common.KitLogger;
+import org.eclipse.jkube.kit.common.util.XMLUtil;
+import org.w3c.dom.Document;
+import org.w3c.dom.Element;
+import org.w3c.dom.Node;
+import org.w3c.dom.NodeList;
+import org.xml.sax.SAXException;
+
+import javax.xml.parsers.ParserConfigurationException;
+import javax.xml.transform.TransformerException;
+import javax.xml.xpath.XPathExpressionException;
+import java.io.File;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+
+public class MigrateService {
+    private File projectBasedir;
+    private static final String POM_XML = "pom.xml";
+    private static final String FABRIC8 = "fabric8";
+    private static final String JKUBE = "jkube";
+    private static final String DEFAULT_FABRIC8_RESOURCE_FRAGMENT_DIRECTORY = "src/main/" + FABRIC8;
+    private static final String DEFAULT_RESOURCE_FRAGMENT_DIRECTORY = "src/main/" + JKUBE;
+    private static final String ARTIFACT_ID = "artifactId";
+    private static final String GROUP_ID = "groupId";
+    private static final String VERSION = "version";
+    private static final String FABRIC8_MAVEN_PLUGIN_ARTIFACT_ID = FABRIC8 + "-maven-plugin";
+    private static final String POM_PROJECT_PATH = "/project";
+    private static final String POM_PROFILES_PATH = POM_PROJECT_PATH + "/profiles";
+    private static final String POM_PLUGINS_PATH = "/build/plugins";
+    private static final String POM_PROPERTIES_PATH = "/properties";
+    private static final String POM_PLUGIN_PATH = "/plugin";
+    private KitLogger logger;
+
+    public MigrateService(File projectBaseDirectory, KitLogger logger) {
+        this.projectBasedir = projectBaseDirectory;
+        this.logger = logger;
+    }
+
+    public void migrate(String pluginGroupId, String pluginArtifactId, String pluginVersion) throws IOException, ParserConfigurationException, SAXException, TransformerException, XPathExpressionException {
+        File pomFile = getPomFile();
+        String pomContent = new String(Files.readAllBytes(pomFile.toPath()), StandardCharsets.UTF_8);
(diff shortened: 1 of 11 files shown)


kubernetes-maven-plugin/plugin/src/main/java/org/eclipse/jkube/maven/plugin/mojo/MigrateMojo.java
```diff
@@ -0,0 +1,46 @@
...
+public class MigrateMojo extends AbstractJKubeMojo {
+    private static final String PLUGIN_ARTIFACT_ID = "kubernetes-maven-plugin";
+    private static final String PLUGIN_GROUP_ID = "org.eclipse.jkube";
+
+    @Override
+    public void executeInternal() throws MojoExecutionException {
+        log = createLogger(null);
+        File projectBasedir = project.getBasedir();
+        try {
+            MigrateService migrateService = new MigrateService(projectBasedir, PLUGIN_GROUP_ID, getPluginArtifactId(), MavenUtil.getVersion(PLUGIN_GROUP_ID, getPluginArtifactId()), log);
```
manusa, line 36: Service shouldn't be instantiated it should be provided by the JKubeServiceHub

jkube-kit/common/src/test/java/org/eclipse/jkube/kit/common/util/XMLUtilTest.java
```diff
@@ -0,0 +1,104 @@
...
+    @Test
+    public void testReadXML() throws URISyntaxException, IOException, SAXException, ParserConfigurationException {
+        // Given
+        File sampleXML = new File(getClass().getResource("/test-project/pom.xml").toURI());
+
+        // When
+        Document document = XMLUtil.readXML(sampleXML);
+
+        // Then
+        assertNotNull(document);
```
manusa, line 49: Add at least a verification to make sure that the unmarshaled document actually matches with the content of the document

jkube-kit/common/src/test/java/org/eclipse/jkube/kit/common/util/XMLUtilTest.java
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    @Test
+    public void testXMLWrite() throws URISyntaxException, IOException, ParserConfigurationException, SAXException, TransformerException, XPathExpressionException {
```
manusa, line 53: // GIven
This test should start by creating a Document with code only
// When
Write the document to a file (single statement)
// Then
Compare the written file to a an expected file

jkube-kit/common/src/main/java/org/eclipse/jkube/kit/common/util/XMLUtil.java
```diff
@@ -0,0 +1,93 @@
...
+        return (Node) xpath.compile(xPathExpression).evaluate(doc, XPathConstants.NODE);
+    }
+
+    public static String getNodeValueFromDocument(Document doc, String xPathExpression) throws XPathExpressionException {
+        XPathFactory xPathfactory = XPathFactory.newInstance();
+        XPath xpath = xPathfactory.newXPath();
+        return (String) xpath.compile(xPathExpression).evaluate(doc, XPathConstants.STRING);
+    }
+
+    public static int findNodeInList(NodeList nodeList, String nodeName, String desiredValue) {
```
manusa, line 81: You can use XPath queries to do this for you

https://www.w3schools.com/xml/xpath_syntax.asp

https://stackoverflow.com/questions/32959791/filtering-xml-document-using-xpath-in-java

manusa:
In general, most of the filtering operations you are performing can be accomplished by XPath, try to use that as it will be more readable and probably more efficient.

jkube-kit/common/src/main/java/org/eclipse/jkube/kit/common/service/MigrateService.java
```diff
@@ -0,0 +1,147 @@
...
+            if (!isRenamed) {
+                logger.warn("Unable to rename resource fragment directory in project");
+            } else {
+                logger.info("Renamed " + DEFAULT_FABRIC8_RESOURCE_FRAGMENT_DIRECTORY + " to " + DEFAULT_RESOURCE_FRAGMENT_DIRECTORY);
+            }
+        }
+    }
+
+    private void modifyFMPPropertiesInsidePom(Document dom) throws XPathExpressionException {
+        Node propertiesNode = XMLUtil.getNodeFromDocument(dom, "/project/properties");
```
dev-gaur, line 133: `"/project/properties"` can also be converted into a constant string.

jkube-kit/common/src/main/java/org/eclipse/jkube/kit/common/util/XMLUtil.java
```diff
@@ -0,0 +1,88 @@
...
+        XPath xpath = xPathfactory.newXPath();
+        return (String) xpath.compile(xPathExpression).evaluate(doc, XPathConstants.STRING);
+    }
+
+    private static DocumentBuilderFactory getDocumentBuilderFactory() throws ParserConfigurationException {
+        DocumentBuilderFactory documentBuilderFactory = DocumentBuilderFactory.newInstance();
+        documentBuilderFactory.setAttribute(XMLConstants.FEATURE_SECURE_PROCESSING, true);
+        documentBuilderFactory.setAttribute(XMLConstants.ACCESS_EXTERNAL_DTD, "");
+        documentBuilderFactory.setAttribute(XMLConstants.ACCESS_EXTERNAL_SCHEMA, "");
+        documentBuilderFactory.setFeature("http://apache.org/xml/features/nonvalidating/load-external-dtd", false);
```
dev-gaur, line 80: These urls can be converted into an array of strings and then we can use a loop for setting them here..

better than hardcoding

PR #244, opened 2020-06-15
Fix #218: Remove RuntimeMode mode parameter from Plugins

### LINKED ISSUE #218: Refactor Build mode
## Description

Does having a `kubernetes` build mode make sense when we have separate plugins (`openshift-maven-plugin` & `kubernetes-maven-plugin`)?

What's the point of having `mvn oc:build `-Djkube.mode=kubernetes` when we can always do `mvn k8s:build` even if we are running on an OpenShift cluster?

rohanKanojia:
## Description

Fix #218 
Removed `jkube.mode` parameter from MOJOs. It won't be computed after this change(earlier default value was `RuntimeMode.auto`). Kubernetes Maven Plugin would send `RuntimeMode.kubernetes` and Openshift Maven Plugin would send `Runtimemode.openshift`.

## Type of change

 - [X] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist


manusa:
Quarkus tests for OpenShift will never pass since they are relying on [Kubernetes mode](https://github.com/jkubeio/jkube-integration-tests/blob/d666125c0a94f5b50c991f75eaed08ab7d19705f/it/src/test/java/org/eclipse/jkube/integrationtests/quarkus/rest/QuarkusOcITCase.java#L77) ([here too](https://github.com/jkubeio/jkube-integration-tests/blob/d666125c0a94f5b50c991f75eaed08ab7d19705f/it/src/test/java/org/eclipse/jkube/integrationtests/quarkus/rest/QuarkusOcITCase.java#L91)).

Try to update the tests so they use the `jkube.build.strategy=docker` property instead. Once tests are updated and merged, the full test suite should be successful.

rohanKanojia:
ah, ok

jkube-kit/config/resource/src/main/java/org/eclipse/jkube/kit/config/resource/RuntimeMode.java
```diff
@@ -33,36 +33,23 @@
      * onto cluster. It can be used both on vanilla Kubernetes and
      * OpenShift.
      */
-    kubernetes(false, "Kubernetes"),
+    KUBERNETES("Kubernetes"),
```
manusa, line 36: Plexus deserialization is not configurable for enums (this is the reason why this values were still in lower-case).
Make sure that this value is never deserialized from a Maven property (i.e. `jkube.mode`), if it still is deserialized somewhere, it will fail).

jkube-kit/config/resource/src/main/java/org/eclipse/jkube/kit/config/resource/RuntimeMode.java
```diff
@@ -33,36 +33,23 @@
...
+    OPENSHIFT("OpenShift");
 
-    /**
-     * Detect automatically whether running cluster is OpenShift or Kuberentes.
-     * This is done by contacting cluster API server.
-     */
-    auto(true, "Auto");
-
-    public static final RuntimeMode DEFAULT = RuntimeMode.auto;
     public static final String FABRIC8_EFFECTIVE_PLATFORM_MODE = "jkube.internal.effective.platform.mode";
```
manusa, line 45: Maybe it's time to rename this constant

jkube-kit/config/service/src/main/java/org/eclipse/jkube/kit/config/service/JKubeServiceHub.java
```diff
@@ -71,16 +70,9 @@ public JKubeServiceHub(
     private void init() {
         Objects.requireNonNull(configuration, "JKubeKitConfiguration is required");
         Objects.requireNonNull(log, "log is a required parameter");
-        if (platformMode == null) {
```
manusa, line 74: If platformMode is now required, add the nullcheck:
`Objects.requireNonNull(platformMode, "platformMode is a required parameter");`

jkube-kit/enricher/generic/src/test/java/org/eclipse/jkube/enricher/generic/openshift/AutoTLSEnricherTest.java
```diff
@@ -91,18 +91,18 @@ private AdaptTestConfig(PlatformMode mode, String initContainerNameConfig, Strin
...
-                new AdaptTestConfig(PlatformMode.openshift, null, "tls-jks-converter", null,
+                new AdaptTestConfig(RuntimeMode.OPENSHIFT, null, "tls-jks-converter", null,
                         "jimmidyson/pemtokeystore:v0.1.0", null, "tls-pem", "jks-b", "jks-b"),
-                new AdaptTestConfig(PlatformMode.openshift, "test-container-name", "test-container-name", "image/123",
+                new AdaptTestConfig(RuntimeMode.OPENSHIFT, "test-container-name", "test-container-name", "image/123",
                         "image/123", "tls-a", "tls-a", "jks-b", "jks-b") };
 
         for (final AdaptTestConfig tc : data) {
-            TreeMap configMap = new TreeMap() {
+            TreeMap<String, String> configMap = new TreeMap<String, String>() {
```
manusa, line 105:
```suggestion
            TreeMap<String, String> configMap = new TreeMap<>() {
```

kubernetes-maven-plugin/it/src/it/simple-with-route-flag-false/invoker.properties
```diff
@@ -13,5 +13,5 @@
 #
 
 invoker.goals.1=clean k8s:resource
-invoker.mavenOpts=-Djkube.verbose -Djkube.mode=openshift -Dfabric8.openshift.generateRoute=false
+invoker.mavenOpts=-Djkube.verbose -Dfabric8.openshift.generateRoute=false
```
manusa, line 16: Probably the second option `-Dfabric8.openshift.generateRoute=false` has no effect (if it does then something is wrong), we should remove it.

kubernetes-maven-plugin/it/src/it/simple-with-route-flag-true/invoker.properties
```diff
@@ -13,5 +13,5 @@
 #
 
 invoker.goals.1=clean k8s:resource
-invoker.mavenOpts=-Djkube.verbose -Djkube.mode=openshift -Dfabric8.openshift.generateRoute=true
-invoker.debug=false
\ No newline at end of file
+invoker.mavenOpts=-Djkube.verbose -Djkube.openshift.generateRoute=true
```
manusa, line 16: https://github.com/eclipse/jkube/pull/244/files#r442657988

kubernetes-maven-plugin/it/src/it/simple-with-route-flag-true/expected/kubernetes.yml
```diff
@@ -84,7 +84,7 @@ items:
             valueFrom:
               fieldRef:
                 fieldPath: metadata.namespace
-          image: jkube-maven-sample-zero-config:latest
+          image: jkube/jkube-maven-sample-zero-config:latest
```
manusa, line 87: Is this because previously this file was ignored?

openshift-maven-plugin/it/src/it/simple-with-route-flag-false/invoker.properties
```diff
@@ -13,5 +13,5 @@
 #
 
 invoker.goals.1=clean oc:resource
-invoker.mavenOpts=-Djkube.verbose -Djkube.mode=openshift -Dfabric8.openshift.generateRoute=false
+invoker.mavenOpts=-Djkube.verbose -Dfabric8.openshift.generateRoute=false
```
manusa, line 16: https://github.com/eclipse/jkube/pull/244/files#r442657988

manusa: Invalid comment, please ignore

openshift-maven-plugin/plugin/src/main/java/org/eclipse/jkube/maven/plugin/mojo/build/OpenshiftResourceMojo.java
```diff
@@ -28,6 +30,48 @@
...
+    private Boolean generateRoute;
+
+    @Parameter(property = "jkube.openshift.enableAutomaticTrigger", defaultValue = "true")
+    private Boolean enableAutomaticTrigger;
+
+    @Parameter(property = "jkube.openshift.imageChangeTrigger", defaultValue = "true")
+    private Boolean enableImageChangeTrigger;
+
+    @Parameter(property = "jkube.openshift.enrichAllWithImageChangeTrigger", defaultValue = "false")
+    private Boolean erichAllWithImageChangeTrigger;
```
manusa, line 73: Are these properties dead? Could you open an issue providing details (if you know about them)?
It seems that at least some of them are used on Enrichers, but are not read from the Mojo.

kubernetes-maven-plugin/it/src/it/simple-with-route-flag-true/expected/kubernetes.yml
rohanKanojia: Actually, earlier this test was running with `jkube.mode=openshift`. With this PR, runtime mode is KUBERNETES for Kubernetes Maven Plugin.
https://github.com/eclipse/jkube/blob/48ccd83fc8ffd81e3d8716b4fa8b43177ae6755d/kubernetes-maven-plugin/it/src/it/simple-with-route-flag-true/invoker.properties#L16

openshift-maven-plugin/plugin/src/main/java/org/eclipse/jkube/maven/plugin/mojo/build/OpenshiftResourceMojo.java
rohanKanojia: Umm, you mean add javadocs. I see these are covered in plugin documentation: https://www.eclipse.org/jkube/docs/openshift-maven-plugin#Supported-Properties-Resource

kubernetes-maven-plugin/it/src/it/simple-with-route-flag-true/expected/kubernetes.yml
manusa: So this file was being ignored previously, right? i.e. if we delete it from master/main branch, tests won't break

rohanKanojia: I don't think this file is ignored.

openshift-maven-plugin/plugin/src/main/java/org/eclipse/jkube/maven/plugin/mojo/build/OpenshiftResourceMojo.java
manusa: Ok, let me rephrase. These properties are dead in the Mojo, they are directly computed/retrieved in the Enricher, meaning these fields can be deleted (fields are ignored, same as provided defaults).
So what we need is an issue providing details about what these properties do (link to docs is good), and what should we do with them. Basically, we need an answer to why are they in the Mojo if they are ignored > this seems like the result of an abandoned work in progress. And what we want, complete their movement to the Mojo (properties no longer retrieved from Enricher) **or** complete their movement to the Enricher (delete them from the Mojo).

rohanKanojia: Thanks for explanation. I've filed https://github.com/eclipse/jkube/issues/259 for this

kubernetes-maven-plugin/it/src/it/simple-with-route-flag-true/expected/kubernetes.yml
manusa: If this is the case, was previous behavior was wrong? Why were `kubernetes.yml` files generated if runtime mode was `openshift`?

rohanKanojia: I think `kubernetes.yml` were getting generated because `kubernetes-maven-plugin` goals were being invoked. This `jkube.mode` property was changing default generator image name depending upon runtime mode. Earlier with `jkube.mode` set to Openshift, they were generating image name for Openshift builds. Right now we have same test in openshift and it's image field looks like this:
https://github.com/eclipse/jkube/blob/b2e25f66df900beede673ad4d236f1f8e900eb0e/openshift-maven-plugin/it/src/it/simple-with-route-flag-true/expected/openshift.yml#L90

jkube-kit/enricher/generic/src/test/java/org/eclipse/jkube/enricher/generic/openshift/AutoTLSEnricherTest.java
rohanKanojia: I tried doing this but I couldn't use '<>' with anonymous inner classes. Our project is at JDK8 right now, maybe we should upgrade to JDK11? WDYT?

manusa: We should not be doing this, this is really bad.
Creating an inner class extending a TreeMap is something I don't see really the purpose here, and is really a bad practice.
Just instantiate the Map and put the entries after.

kubernetes-maven-plugin/it/src/it/simple-with-route-flag-true/expected/kubernetes.yml
manusa: > Right now we have same test in openshift and it's image field looks like this:

Yes, but it's generated in an `openshift.yml` file (as expected).

The fact that this wasn't the case before, and that instead of generating an `openshift.yml` file a `kubernetes.yml` file was being generated with openshift specific configurations (i.e. `image`)  and not even completed with deploymentconfig files, makes me understand that previous behavior was wrong.

So conclusion is that this file was not being ignored before due to an inconsistent behavior, but after the current PR, things will behave as expected and only specific cluster files will be generated with the **right** configs.

rohanKanojia: cool, thanks. I would mark this as resolved.

Issue #207, opened 2020-05-21
Increase Unit test coverage for jkube-kit/common (achieve 45% code coverage)


anandrkskd:
# Description
Increase current unit test coverage for the following directory https://github.com/eclipse/jkube/tree/master/jkube-kit/common 

# Acceptance Criteria
- [ ] 45% Code coverage for the [specified module](https://sonarcloud.io/component_measures?id=jkubeio_jkube&metric=coverage&selected=jkubeio_jkube%3Ajkube-kit%2Fcommon%2Fsrc)
- [ ] It should cover complex scenario
- [ ] Scenario test should be passed, otherwise, issues will re-reported

/kind user-story

manusa:
https://sonarcloud.io/component_measures?id=jkubeio_jkube&metric=coverage&selected=jkubeio_jkube%3Ajkube-kit%2Fcommon

manusa:
outdated, no longer applicable. Also, current metric is at 56%.

PR #247, opened 2020-06-17
Unit test for Openshifthelper

anandrkskd:
## Description
Increase coverage for OpenShiftHelper in scope of #207 

## Type of change

 - [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist
 - [ ] ~I Added [CHANGELOG](../CHANGELOG.md) entry~
 - [ ] ~I have updated the [documentation](../kubernetes-maven-plugin/doc) accordingly~
 - [ ] ~I tested my code in Kubernetes~
 - [ ] ~I tested my code in OpenShift~

### DIFF
--- a/jkube-kit/common/src/test/java/org/eclipse/jkube/kit/common/util/OpenshiftHelperTest.java
+++ b/jkube-kit/common/src/test/java/org/eclipse/jkube/kit/common/util/OpenshiftHelperTest.java
@@ -1,0 +1,307 @@
+/**
+ * Copyright (c) 2019 Red Hat, Inc.
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at:
+ *
+ *     https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *   Red Hat, Inc. - initial API and implementation
+ */
+package org.eclipse.jkube.kit.common.util;
+
+import io.fabric8.kubernetes.api.model.KubernetesList;
+import io.fabric8.kubernetes.api.model.Pod;
+import io.fabric8.kubernetes.api.model.PodBuilder;
+import io.fabric8.kubernetes.client.KubernetesClient;
+import io.fabric8.kubernetes.client.KubernetesClientException;
+import io.fabric8.openshift.api.model.Template;
+import io.fabric8.openshift.api.model.TemplateBuilder;
+import io.fabric8.openshift.client.OpenShiftClient;
+
+import mockit.Expectations;
+import mockit.Mocked;
+import org.junit.Test;
+
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+
+public class OpenshiftHelperTest {
+
+    @Test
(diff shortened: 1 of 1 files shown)


jkube-kit/common/src/test/java/org/eclipse/jkube/kit/common/util/OpenshiftHelperTest.java
```diff
@@ -0,0 +1,312 @@
...
+import static org.junit.Assert.assertNotEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+
+public class OpenshiftHelperTest {
+
+    @Rule
+    public KubernetesServer server = new KubernetesServer();
+//    OpenShiftMockServer mockServer = new OpenShiftMockServer(false);
```
manusa, line 44: Remove dead code

manusa:
The real purpose here would be to create **Unit tests**, as of now you are creating _Integration Tests_ since many of the tested parts + mocks are actually in Kuberentes-Client module.

It would be a better approach to Mock KuberentesClient class where necessary.

The main drawback of an integration test like this one is that if things change in Kubernetes Client (e.g. API endpoints hit to determine which type of client it is), our tests will start failing with no apparent reason. When implementing a unit test, what we want is to mock any public API method from our dependencies so that what we test is actually our code and not our dependency code.

manusa:
As an example of the previous approach, test `testOpenShiftClientWithOpenShift` can be rewritten to 2 tests that will test all conditions:
```java
    @Test
    public void testOpenShiftClientWithAdaptableToOpenShift(@Mocked KubernetesClient kc, @Mocked OpenShiftClient oc) {
        // Given
        // @formatter:off
        new Expectations() {{
            kc.adapt(OpenShiftClient.class);
            result = oc;
        }};
        // @formatter:on
        //When
        OpenShiftClient result = OpenshiftHelper.asOpenShiftClient(kc);
        //Then
        assertNotNull(result);
    }

    @Test
    public void testOpenShiftClientWithOpenShift(@Mocked OpenShiftClient oc) {
        //When
        OpenShiftClient result = OpenshiftHelper.asOpenShiftClient(oc);
        //Then
        assertNotNull(result);
    }
```

You don't really care about the internals of KubernetesClient but what the public API of KubernetesClient returns. So in case any of the internals of KubernetesClient (or any other library) change in the future, this test will still be reliable and won't need changes. i.e. Test will only require changes if KubernetesClient public API changes.

jkube-kit/common/pom.xml
```diff
@@ -135,6 +135,12 @@
       <groupId>com.networknt</groupId>
       <artifactId>json-schema-validator</artifactId>
     </dependency>
+      <dependency>
+          <groupId>io.fabric8</groupId>
+          <artifactId>kubernetes-server-mock</artifactId>
+          <version>4.10.2</version>
```
rohanKanojia, line 141: I don't think you need to specify version here since it's covered by `<dependencyManagement>`

rohanKanojia: anyway, If you are not going to mock Kubernetes API. You can remove this

jkube-kit/common/src/test/java/org/eclipse/jkube/kit/common/util/OpenshiftHelperTest.java
```diff
@@ -0,0 +1,307 @@
...
+
+public class OpenshiftHelperTest {
+
+    @Rule
+    public KubernetesServer server = new KubernetesServer();
+
+    @Test
+    public void testAsOpenShiftClientWithNoOpenShift() {
+        // Given
+        KubernetesClient client = server.getClient();
```
rohanKanojia, line 49: I think you should be able to use mocked `KubernetesClient` here also.

Issue #283, opened 2020-07-08
Add support for WildFly Bootable JAR


jfdenise:
We are introducing a new way to run WildFly applications inside a bootable JAR starting WildFly 20. The maven plugin is defined there: https://github.com/wildfly-extras/wildfly-jar-maven-plugin

This plugin relies on Galleon to identify the server capabilities a user wants to see provisioned.

We have a need for:
* A new generator (that extends java-exec) and disable both Prometheus and Jolokia.
* A new HealthcheckEnricher to discover and configure Probes.

A WIP branch to highlight early work: https://github.com/jfdenise/jkube/pull/new/wildfly-jar

PR #294, opened 2020-07-17
Fix for issue #283, Add support for WildFly Bootable JAR

jfdenise:
Issue: https://github.com/eclipse/jkube/issues/283
Signed-off-by: JF Denise <jdenise@redhat.com>

WARNING: That is a draft PR to acquire feedback on the approach. No integration tests yet. I have some questions.

## Description
We are adding a generator and enricher for WildFly Bootable JAR. The requirements are for now quite limited (and similar to Thorntail V2):
* Discovery of the org.wildfly.plugins:wildfly-jar-maven-plugin Maven plugin.
* A new generator that extends java-exec. We are disabling jolokia and prometheus agent for now. We hope to fix the issue we currently have in a near future.
* An new enricher that adds heath =checks (readiness and liveness).

## Questions
* Our generator doesn't add "extra Java Options" although we need -Djava.net.preferIPv4Stack=true. The user has to set  it explicitly. So we would have to document it. I tried to add it to our Generator extra java options but a user re-defining JAVA_OPTIONS with his own options would override extra options added by the Generator and loose it. Am I correct there? What is the best approach for mandatory Java options we want to see present whatever the user Java options?

## Type of change

## Checklist
 - [X] No new bugs, code smells, etc. in [SonarCloud](https://sonarcloud.io/dashboard?id=jkubeio_jkube) report

NOT YET integration tests in https://github.com/jkubeio/jkube-integration-tests


manusa:
> ## Questions
> * Our generator doesn't add "extra Java Options" although we need -Djava.net.preferIPv4Stack=true. The user has to set  it explicitly. So we would have to document it. I tried to add it to our Generator extra java options but a user re-defining JAVA_OPTIONS with his own options would override extra options added by the Generator and loose it. Am I correct there? What is the best approach for mandatory Java options we want to see present whatever the user Java options?

I really haven't tried what happens if `JAVA_OPTIONS` env variable is provided by the user too. The WildFlySwarmGenerator is using [this strategy](https://github.com/eclipse/jkube/blob/8555e05f44ae6e7e0185bf7473c2c8387f516a7d/jkube-kit/generator/wildfly-swarm/src/main/java/org/eclipse/jkube/generator/wildflyswarm/WildFlySwarmGenerator.java#L53) to set the flag you mention (Vert.x generator [too](https://github.com/eclipse/jkube/blob/8555e05f44ae6e7e0185bf7473c2c8387f516a7d/jkube-kit/jkube-kit-vertx/src/main/java/org/eclipse/jkube/vertx/generator/VertxGenerator.java#L58)).

As I said, I haven't tried this personally (but will). If options are overridden by user providing extra env variables, then some fix should be performed [here](https://github.com/eclipse/jkube/blob/8555e05f44ae6e7e0185bf7473c2c8387f516a7d/jkube-kit/generator/java-exec/src/main/java/org/eclipse/jkube/generator/javaexec/JavaExecGenerator.java#L161). Anyway, this should be reported as a separate bug and fix because other generators are affected by this.

manusa:
After a deeper check of PropertyConfigHandler, and the generator processes, I'm not even sure that the user can override the env variables provided by any generator.
I think the only way to provide [env variables](http://localhost:63342/jkube/kubernetes-maven-plugin-doc/target/generated-docs/index.html?_ijt=5j8bqhr5idbj9i3r5m8nkcd5ll#misc-env) for a build is by providing a complete ImageConfiguration, which will then render the Generator useless.

jfdenise:
@manusa @rohanKanojia , seems that the remaining failing checks are unrelated to this change.

manusa:
> manusa rohanKanojia , seems that the remaining failing checks are unrelated to this change.

All checks are green now ✔️ 👏

openshift-maven-plugin/plugin/src/main/resources/META-INF/jkube/profiles-default.yml
```diff
@@ -102,7 +104,7 @@
     includes:
     - spring-boot
     - wildfly-swarm
-    - thorntail-v2
+    - wildfly-jar
```
manusa, line 107: Is the substitution/replacement intended?

quickstarts/maven/wildfly-jar/pom.xml
```diff
@@ -0,0 +1,146 @@
...
+
+  <profiles>
+    <profile>
+      <id>kubernetes</id>
+      <build>
+	<plugins>
+	  <plugin>
+	    <groupId>org.eclipse.jkube</groupId>
+	    <artifactId>kubernetes-maven-plugin</artifactId>
+            <version>1.0.0-SNAPSHOT</version>
```
manusa, line 83: Please bind this to ${project.version} to ease the maintenance process (same for line 117). 
Project version (line 23) can be set to `1.0.0-SNAPSHOT` since this won't work in current latest release (1.0.0-alpha-4).

quickstarts/maven/wildfly-jar/pom.xml
```diff
@@ -0,0 +1,146 @@
...
+
+  <properties>
+    <version.wildfly.jar>2.0.0.Alpha6</version.wildfly.jar>
+    <version.wildfly>20.0.0.Final</version.wildfly>
+    <maven.compiler.source>1.8</maven.compiler.source>
+    <maven.compiler.target>1.8</maven.compiler.target>
+    <failOnMissingWebXml>false</failOnMissingWebXml>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <!-- crash image name -->
+    <!--<jkube.generator.name>%a/%g:%l</jkube.generator.name>-->
```
manusa, line 35: Is this showcasing something or is it dead code?

openshift-maven-plugin/plugin/src/main/resources/META-INF/jkube/profiles-default.yml
jfdenise: Was not intended, thank-you for catching it!

quickstarts/maven/wildfly-jar/pom.xml
jfdenise: Ok, fixed.

quickstarts/maven/wildfly-jar/pom.xml
jfdenise: Dead code, removed.

manusa:
BTW, there are conflicts with the master branch. Documentation for Enricher entries were moved to separate files.

jfdenise:
@manusa , fixing the merge issue.

jfdenise:
@manusa , merge conflict fixed.

PR #349, opened 2020-08-07
Fix #341 : ImageChangeTrigger now gets applied when resource fragments used

### LINKED ISSUE #341: OpenShift Maven Plugin doesn't add ImageChange triggers when merging from a deployment fragment
### Description

Eclipse JKube is not adding ImageChange triggers to `DeploymentConfig` due to which applications deployed in openshift are failing with `ImagePullErr`. This seems to happen in a case when project has some deployment fragment in `src/main/jkube` directory. I have to explicitly add `jkube.openshift.enrichAllWithImageChangeTrigger` property in order to add `ImageChange` triggers:
```
    triggers:
    - type: ConfigChange
    - imageChangeParams:
        automatic: true
        containerNames:
        - spring-boot
        from:
          kind: ImageStreamTag
          name: spring-boot-camel:latest
      type: ImageChange
```

Same project works fine without any extra configuration with Fabric8 Maven Plugin

### Info

* Eclipse JKube version : 1.0.0-rc-1
* Maven version (`mvn -v`) :
```
Apache Maven 3.6.3 (cecedd343002696d0abb50b32b541b8a6ba2883f)
Maven home: /opt/apache-maven
Java version: 1.8.0_251, vendor: Oracle Corporation, runtime: /usr/java/jdk1.8.0_251-amd64/jre
OS name: "linux", version: "5.6.19-300.fc32.x86_64", arch: "amd64", family: "unix"
```
* Kubernetes / Red Hat OpenShift setup and version : OpenShift 4.5.1

- Clone https://github.com/rohanKanojia/spring-boot-camel
- Disable `jkube.openshift.enrichAllWithImageChangeTrigger` property
- deploy application using `mvn clean install -DskipTests oc:deploy -Popenshift`
- application would fail to deploy with `ImagePullErr`

* Sample Reproducer Project : https://github.com/rohanKanojia/spring-boot-camel

rohanKanojia:
Fix #341 

Earlier due to sidecar refactor, JKube wasn't able to find out
default application container's name for which we need to add
ImageChange triggers during openshift S2I build when we were
providing resource fragments in which container name was absent.

## Type of change

## Checklist

### DIFF
--- a/jkube-kit/enricher/api/src/main/java/org/eclipse/jkube/kit/enricher/api/util/KubernetesResourceUtil.java
+++ b/jkube-kit/enricher/api/src/main/java/org/eclipse/jkube/kit/enricher/api/util/KubernetesResourceUtil.java
@@ -637,10 +637,16 @@ public class KubernetesResourceUtil {
                             container = new Container();
                             containers.add(container);
                         }
-                        // If default container name is not set, add first found
-                        // container as default application container.
-                        if (defaultApplicationContainerName == null) {
+                    }
+                    // If default container name is not set, add first found
+                    // container as default application container from resource
+                    // fragment, if not present set default application container
+                    // name from JKube generated PodSpec
+                    if (defaultApplicationContainerName == null) {
+                        if (container.getName() != null) { // Pick from fragment
                             defaultApplicationContainerName = container.getName();
+                        } else if (defaultContainer.getName() != null) { // Pick from default opinionated PodSpec
+                            defaultApplicationContainerName = defaultContainer.getName();
                         }
                     }
 
@@ -1002,4 +1008,3 @@ public class KubernetesResourceUtil {
         return true;
     }
 }
-
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ Usage:
 * Fix #332: OpenShift build resources are deleted (as long as build config manifest is available)
 * Fix #350: Prevents default docker configuration overwriting XML assembly configuration
 * Fix #340: Exclude the main artifact from Docker build when Fat Jar is detected (JavaExecGenerator)
+* Fix #341: JKube doesn't add ImageChange triggers in DC when merging from a deployment fragment 
 
 ### 1.0.0-rc-1 (2020-07-23)
 * Fix #252: Replace Quarkus Native Base image with ubi-minimal (same as in `Dockerfile.native`)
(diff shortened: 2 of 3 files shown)


jkube-kit/enricher/api/src/main/java/org/eclipse/jkube/kit/enricher/api/util/KubernetesResourceUtil.java
```diff
@@ -650,76 +650,115 @@ public static String mergePodSpec(PodSpecBuilder builder, PodSpec defaultPodSpec
         int size = defaultContainers.size();
         if (size > 0) {
             if (containers == null || containers.isEmpty()) {
+                defaultApplicationContainerName = defaultContainers.get(0).getName();
```
dev-gaur, line 653: Do we always the guarantee that the 0th container is the app container? I think we should add a validation here to keep this logic robust.

Please correct me if I'm wrong.

rohanKanojia: I think you have a point here. During zero-configuration mode, JKube adds only one container(that's why I assumed it would always work). But if you use XML configuration for ImageConfig, you can add as many containers as you wish and there is no way to specify which one is main application container.

dev-gaur: exactly

manusa: With all the code changes, I just realized that this diverges a lot from the initial [proposed solution](https://github.com/eclipse/jkube/issues/341#issuecomment-669915655) (I thought that code was only moved around).

Could we stick to the initial plan (implement simple solution and add the the tests)? This class needs a deep refactor, but with the 1.0 release ahead, it's not the right time to do this.

manusa: After a deeper review, the string returned by this function is absolutely useless and only used as a flag in:

https://github.com/eclipse/jkube/blob/89f36498ab7e2dcc3d7d233e0ae919f4f05fe451/jkube-kit/enricher/generic/src/main/java/org/eclipse/jkube/enricher/generic/ControllerViaPluginConfigurationEnricher.java#L123-L128

 It doesn't really matter if the name is the one for the first container, the second, or anything else `"random"`. 

We really need to refactor all of the related methods after the 1.0.0 release

Issue #385, opened 2020-09-17
WildFly Bootable JAR - Add native support for slim Bootable JAR


jfdenise:
### Description
Thanks to JKube workflow, we have the opportunity to benefit from slim Bootable JAR (a Bootable JAR that doesn't contain JBoss modules artifacts) feature. slim Bootable JAR starts fast, this wil lreduce the server startup time.

The WildFlyJar Generator can detect the path to the generated maven repo cache configured in the wildfly-jar-maven-plugin and installs it in the Docker image in /deployments/[directory name].

In addition, the generator adds the -Dmaven.local.repo=/deployments/[directory name] in the JAVA_OPTIONS. That is required for the server to retrieve JBoss modules artifacts. NB: In order for this part to work properly we need a fix for https://github.com/eclipse/jkube/issues/384

### Info

* Eclipse JKube version :1.0.0

manusa:
Hi @jfdenise 
#384 has been closed via #395.
Besides the aforementioned option, it would be nice to include the `-Djava.net.preferIPv4Stack=true` option too, so that project using the plugin don't need to manually add this.

manusa:
Implemented in #398

PR #398, opened 2020-09-23
Fix for issue #385, WildFly Bootable JAR - Add native support for slim Bootable JAR

jfdenise:
Signed-off-by: JF Denise <jdenise@redhat.com>

The quickstart evolution is tracked by https://github.com/eclipse/jkube/issues/399

## Description
* Support for slim server in WildFly bootable JAR.
* In addition the preferIPV4 is now added by default.
* Upgraded WildFly and wildfly-jar plugin versions in quickstart.

## Type of change

## Checklist

### DIFF
--- a/jkube-kit/jkube-kit-wildfly-jar/src/test/java/org/eclipse/jkube/wildfly/jar/generator/WildflyJARGeneratorTest.java
+++ b/jkube-kit/jkube-kit-wildfly-jar/src/test/java/org/eclipse/jkube/wildfly/jar/generator/WildflyJARGeneratorTest.java
@@ -13,6 +13,7 @@
  */
 package org.eclipse.jkube.wildfly.jar.generator;
 
+import java.io.File;
 import org.eclipse.jkube.generator.api.GeneratorContext;
 import org.eclipse.jkube.kit.config.image.ImageConfiguration;
 import mockit.Expectations;
@@ -22,13 +23,25 @@ import org.junit.Test;
 
 import java.io.IOException;
 import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import org.eclipse.jkube.kit.common.AssemblyFileSet;
+import org.eclipse.jkube.kit.common.Plugin;
+import org.eclipse.jkube.kit.config.resource.ProcessorConfig;
+import org.eclipse.jkube.wildfly.jar.enricher.WildflyJARHealthCheckEnricher;
+import static org.eclipse.jkube.wildfly.jar.generator.WildflyJARGenerator.JBOSS_MAVEN_DIST;
+import static org.eclipse.jkube.wildfly.jar.generator.WildflyJARGenerator.JBOSS_MAVEN_REPO;
+import static org.eclipse.jkube.wildfly.jar.generator.WildflyJARGenerator.PLUGIN_OPTIONS;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertThrows;
+import static org.junit.Assert.fail;
 
 /**
  * @author roland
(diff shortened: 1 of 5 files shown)


jfdenise:
@manusa , the integration tests are failing because the -Djava.net.preferIPv4Stack=true is added twice. It is in the test project fragment and now added by the WildFly JAR generator. If the user defines an option, shouldn't it take precedence on the default ones (added by the generator)?

manusa:
I'm not sure what happens if 
```
java  -Djava.net.preferIPv4Stack=true  -Djava.net.preferIPv4Stack=false...
```
is invoked. Not sure if the first argument or the second will be the "winning".

If the sure wants to override the generated value for JAVA_OPTIONS, he/she may configure this by the `jkube.enricher.jkube-container-env-java-options.disable`option. [As documented](https://github.com/eclipse/jkube/blob/master/kubernetes-maven-plugin/doc/src/main/asciidoc/inc/enricher/_jkube_container_env_java_options.adoc).

We could have added the additional complexity of checking individual options in the new enricher. But IMHO this adds lot's of complexity which are probably unneeded. If the user wants to provide overridden values for the JAVA_OPTIONS in the generator, he/she may as well completely disable the merge functionality and provide the complete list of options.

manusa:
Regarding the tests, we can update them once this gets merged.

manusa:
> @manusa , I want to update the quickstart with this new feature but quickstarts are using version JKube 1.0.0. Could we have the quickstarts to reference the SNAPSHOT build and make the change to the quickstarts when releasing? With 1.0.0, the quickstart is broken.

Yes, this is what we've been doing up until now.

jfdenise:
@manusa , thank-you. The last option on the command line wins. With respect to the quickstart being broken with 1.0.0, is it fine that I update it with the new feature in this PR? Should I postpone it until you release 1.0.1 then open a new PR just for the quickstart?

jfdenise:
For the 2 code smells:
1) Do I need to break into a sequence of if? I personally prefer the way it is written now.
2) I can't use a Checked Exception, the method I am overriding doesn't define an Exception in its signature.

manusa:
> The last option on the command line wins
That's what I thought. So in the merge strategy, whatever is added in the configuration is appended to what the generator adds by default.

> Should I postpone it until you release 1.0.1 then open a new PR just for the quickstart?
Probably that's a good idea now that we have 1.0.0 out. If users want to check out the wildfly-jar example and have a bad experience, they'll probably won't come back.

jfdenise:
OK, then I will postpone it. BTW: In the WildFly JAR project, we expect people wanting to use examples to checkout a release tag. In the tag, the version of the plugin is the released one. In the master branch, all examples depend on the SNAPSHOT. So a user would have to build the plugin if they want to use the examples with the latest/greatest plugin.

manusa:
> For the 2 code smells:
> 
> 1. Do I need to break into a sequence of if? I personally prefer the way it is written now.
> 2. I can't use a Checked Exception, the method I am overriding doesn't define an Exception in its signature.

Number one you can easily reduce complexity by using:
```
final String path = Optional.ofNullable(plugin).map(Plugin::getConfiguration).map(c -> config.get(PLUGIN_OPTIONS)/*...*/.orElse(null);
if (StringUtils.isNotEmpty(path) {
cache = Paths.get(path);
}
```
Or even:
```
Path cache = Optional.ofNullable(plugin).map(Plugin::getConfiguration).map(c -> config.get(PLUGIN_OPTIONS)/*...*/.map(Paths::get).orElse(null);
```

The second is about using a specific Exception. I think we currently don't have any JKube specific RuntimeException, we'll take care of that in future refactors (this code smell is all around the project).

manusa:
> OK, then I will postpone it. BTW: In the WildFly JAR project, we expect people wanting to use examples to checkout a release tag. In the tag, the version of the plugin is the released one. In the master branch, all examples depend on the SNAPSHOT. So a user would have to build the plugin if they want to use the examples with the latest/greatest plugin.

That's an option to consider. The problem with that is that you can't add examples that work on previous versions after the project tag has been released.

Anyway, we're working on that. What we want is a tighter integration with https://www.eclipse.org/jkube/quickstarts/ so probably adding a download link, etc. But this is still in an early WIP stage.

jfdenise:
@manusa , I followed your suggestion and used Optional (although the amount of Cognitive effort for an old developer like me is much bigger ;-))

jfdenise:
@I am changing the state of this draft PR to PR ready for review. 
I will open a PR in integration tests to remove the Java System property in the fragment.

jkube-kit/jkube-kit-wildfly-jar/src/test/java/org/eclipse/jkube/wildfly/jar/generator/WildflyJARGeneratorTest.java
```diff
@@ -53,7 +64,204 @@ public void getEnv() throws IOException {
...
+            boolean failed = true;
+            try {
+                List<AssemblyFileSet> files = generator.addAdditionalFiles();
+                failed = false;
+            } catch(Exception ex) {
+                // XXX Expected
+            }
+            if (!failed) {
+                throw new Exception("Test should have failed, no directory for maven repo");
+            }
```
manusa, line 171:
```suggestion

                List<AssemblyFileSet> files = generator.addAdditionalFiles();
                failed = false;
            } catch(Exception ex) {
                // XXX Expected
            }
            if (!failed) {
                throw new Exception("Test should have failed, no directory for maven repo");
            }
```

manusa, line 171:
```suggestion
            /* Exception result = */assertThrows(Exception.class, () -> {
              generator.addAdditionalFiles();
              fail("Test should have failed, no directory for maven repo");
            });
            // Optionally assert exception e.g. assertEquals("Expected Message", result.getMessage());
```

jfdenise: @manusa , thank-you, I integrated your suggestion.

manusa (approved):
LGTM, thx!!!

rohanKanojia (approved):

manusa:
> I am changing the state of this draft PR to PR ready for review.
> I will open a PR in integration tests to remove the Java System property in the fragment.


Hi @jfdenise do you want us to update the tests or do you have already something?

jfdenise:
Hi @manusa , I just opened https://github.com/jkubeio/jkube-integration-tests/pull/67

manusa:
Great, thank you!!

PR #454, opened 2020-10-29
Fix #146: ApplyMojo should use ApplyService in JKubeServiceHub

### LINKED ISSUE #146: Apply Mojo should use ApplyService in JKubeServiceHub
## Description

Currently ApplyMojo creates its own `ApplyService` instance.

All Mojos should use instance of the services provided by the `JKubeServiceHub` (see BuildMojo)

Relates to: #136

rohanKanojia:
## Description

Fix #146: ApplyMojo now uses ApplyService from JKubeServiceHub

+ Moved applyEntities logic as static methods in ApplyService in order to make ApplyMojo slim

## Type of change

## Checklist

### DIFF
--- a/jkube-kit/config/service/src/main/java/org/eclipse/jkube/kit/config/service/ApplyService.java
+++ b/jkube-kit/config/service/src/main/java/org/eclipse/jkube/kit/config/service/ApplyService.java
@@ -1413,6 +1413,8 @@ public class ApplyService {
     public void applyEntities(String fileName, Set<HasMetadata> entities, KitLogger serviceLogger,
                                  long serviceUrlWaitTimeSeconds, ResourceConfig resources, File resourceDir,
                                  String environment) throws Exception {
+        // Apply Namespace/Project first
+        applyNamespaceOrProjectIfPresent(entities);
         // Apply all items
         applyStandardEntities(fileName, entities);
         logExposeServiceUrl(entities, serviceLogger, serviceUrlWaitTimeSeconds);
--- a/jkube-kit/config/service/src/main/java/org/eclipse/jkube/kit/config/service/ApplyService.java
+++ b/jkube-kit/config/service/src/main/java/org/eclipse/jkube/kit/config/service/ApplyService.java
@@ -58,6 +58,9 @@ import org.eclipse.jkube.kit.common.util.KubernetesHelper;
 import org.eclipse.jkube.kit.common.util.OpenshiftHelper;
 import org.eclipse.jkube.kit.common.util.ResourceUtil;
 import org.eclipse.jkube.kit.common.util.UserConfigurationCompare;
+import org.eclipse.jkube.kit.config.access.ClusterAccess;
+import org.eclipse.jkube.kit.config.resource.JKubeAnnotations;
+import org.eclipse.jkube.kit.config.resource.ResourceConfig;
 import org.eclipse.jkube.kit.config.service.kubernetes.KubernetesClientUtil;
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.lang3.StringUtils;
@@ -67,6 +70,7 @@ import java.io.IOException;
 import java.net.HttpURLConnection;
 import java.nio.charset.Charset;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
@@ -268,10 +272,10 @@ public class ApplyService {
             return;
         }
         if (!isProcessTemplatesLocally()) {
-            String namespace = getNamespace();
+            String currentNamespace = getNamespace();
             String id = getName(entity);
             Objects.requireNonNull(id, "No name for " + entity + " " + sourceName);
-            Template old = openShiftClient.templates().inNamespace(namespace).withName(id).get();
+            Template old = openShiftClient.templates().inNamespace(currentNamespace).withName(id).get();
             if (isRunning(old)) {
                 if (UserConfigurationCompare.configEqual(entity, old)) {
                     log.info("Template has not changed so not doing anything");
@@ -280,12 +284,12 @@ public class ApplyService {
                     // TODO seems you can't update templates right now
                     recreateMode = true;
                     if (recreateMode) {
-                        openShiftClient.templates().inNamespace(namespace).withName(id).delete();
-                        doCreateTemplate(entity, namespace, sourceName);
+                        openShiftClient.templates().inNamespace(currentNamespace).withName(id).delete();
+                        doCreateTemplate(entity, currentNamespace, sourceName);
                     } else {
                         log.info("Updating a Template from %s", sourceName);
                         try {
-                            Object answer = openShiftClient.templates().inNamespace(namespace).withName(id).replace(entity);
+                            Object answer = openShiftClient.templates().inNamespace(currentNamespace).withName(id).replace(entity);
                             log.info("Updated Template: " + answer);
                         } catch (Exception e) {
                             onApplyError("Failed to update Template from " + sourceName + ". " + e + ". " + entity, e);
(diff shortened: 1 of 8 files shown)


manusa:
In order to ease review, have you added/modified/removed any of the logic from the Mojo, or just moved it to the Service class and improved coverage?

rohanKanojia:
There were already some tests added in https://github.com/eclipse/jkube/pull/452/ regarding this PR. I think maybe due to this reason coverage increased after rebasing with master

jkube-kit/config/service/src/main/java/org/eclipse/jkube/kit/config/service/ApplyService.java
```diff
@@ -307,6 +311,10 @@ public OpenShiftClient getOpenShiftClient() {
         return OpenshiftHelper.asOpenShiftClient(kubernetesClient);
     }
 
+    public KubernetesClient getKubernetesClient() {
+        return kubernetesClient;
+    }
```
manusa, line 316: Why would this be exposed?

jkube-kit/config/service/src/main/java/org/eclipse/jkube/kit/config/service/ApplyService.java
```diff
@@ -1390,4 +1398,55 @@ public void processCustomEntities(List<String> customResourceDefinitions, File r
...
+                                 long serviceUrlWaitTimeSeconds, ResourceConfig resources, File resourceDir,
+                                 String environment) throws Exception {
+        // Apply all items
+        applyStandardEntities(applyService, fileName, entities);
+        logExposeServiceUrl(applyService, entities, serviceLogger, serviceUrlWaitTimeSeconds);
+        applyService.processCustomEntities(resources != null ? resources.getCustomResourceDefinitions() : null,
+                resourceDir, environment, resources != null ? resources.getRemotes() : null);
+    }
+
+    private static void logExposeServiceUrl(ApplyService applyService, Set<HasMetadata> entities, KitLogger serviceLogger, long serviceUrlWaitTimeSeconds) throws InterruptedException {
```
manusa, line 1428: This has no sense.
You have a private static method where you provide an instance of the containing class? Just remove the static modifier then.

jkube-kit/config/service/src/main/java/org/eclipse/jkube/kit/config/service/ApplyService.java
```diff
@@ -1390,4 +1398,55 @@ public void processCustomEntities(List<String> customResourceDefinitions, File r
...
+                Namespace ns = (Namespace) projectOrNamespace;
+                applyService.applyNamespace(ns);
+            } else if (projectOrNamespace instanceof Project) {
+                Project project = (Project) projectOrNamespace;
+                applyService.applyProject(project);
+            }
+        }
+    }
+
+    public static void applyEntities(ApplyService applyService, String fileName, Set<HasMetadata> entities, KitLogger serviceLogger,
```
manusa, line 1418: Take this comment and apply it here https://github.com/eclipse/jkube/pull/454/files#r524302202

If you pass an instance of the enclosing class, then it really makes no sense to make this method static

jkube-kit/config/service/src/main/java/org/eclipse/jkube/kit/config/service/ApplyService.java
```diff
@@ -1390,4 +1394,55 @@ public void processCustomEntities(List<String> customResourceDefinitions, File r
             }
         }
     }
+
+    public static void applyNamespaceOrProjectIfPresent(ApplyService applyService, Collection<HasMetadata> entities) {
```
manusa, line 1398: Same as the others. Static method which accepts an instance of the enclosing class.
>> Remove static modifier and ApplyService instance parameter

jkube-kit/config/service/src/main/java/org/eclipse/jkube/kit/config/service/ApplyService.java
```diff
@@ -1390,4 +1394,55 @@ public void processCustomEntities(List<String> customResourceDefinitions, File r
             }
         }
     }
+
+    public void applyNamespaceOrProjectIfPresent(Collection<HasMetadata> entities) {
+        List<HasMetadata> projectOrNamespaceList = entities.stream()
+                .filter(e -> e instanceof Namespace || e instanceof Project)
+                .collect(Collectors.toList());
+
+        for (HasMetadata projectOrNamespace : projectOrNamespaceList) {
```
manusa, line 1403: why not do a `forEach` here.
You are generating a list of HasMetadata, and then re-iterating over that list.

kubernetes-maven-plugin/plugin/src/main/java/org/eclipse/jkube/maven/plugin/mojo/build/ApplyMojo.java
```diff
@@ -186,55 +178,18 @@ public void executeInternal() throws MojoExecutionException {
...
-                    entities.remove(entity);
-                    break;
-                }
-            }
-
-            if (!namespaceEntityExist) {
-                applyService.applyNamespace(namespace);
-            }
-
+            applyService.applyNamespaceOrProjectIfPresent(entities);
```
manusa, line 188: This is weird and unclear.
Could you explain what's going on?
We retrieve the namespace, but the namespace is only set/configured later on in the method.
In addition, the invocation in line 188 (`applyService.applyNamespaceOrProjectIfPresent`) suggests we are already doing something here, but we only configure the namespace later on.
There's also a call `configureApplyService` in line 181. It feels that the namespace part should be really configured there.

My understanding is that if the namespace is configured later, is because it has something to do with the processing and it affects the end result. 
If this is the case `setNamespace` should be completely removed and shifted to an argument of any method that performs processing.

rohanKanojia: I think this code handles the case when there is a `Namespace` present in `entities`. I think It handles the case when you want your resources to be created in a new namespace which would be created by jkube during apply phase. Namespace gets created first, then resources are applied in that particular namespace.

okay, let me shift `setNamespace` to `configureApplyService`

kubernetes-maven-plugin/plugin/src/main/java/org/eclipse/jkube/maven/plugin/mojo/build/ApplyMojo.java
```diff
@@ -184,58 +176,15 @@ public void executeInternal() throws MojoExecutionException {
...
-                }
-            }
-
-            if (!namespaceEntityExist) {
-                applyService.applyNamespace(namespace);
-            }
+            configureApplyService(kubernetes, entities);
 
-            applyService.setNamespace(namespace);
+            applyService.applyNamespaceOrProjectIfPresent(entities);
```
manusa, line 184: Sorry, but I'm still unsure why this method is not part of `applyService#applyEntities` logic.
Why do we have to call it separately?
Is it something that applies specifically to Maven projects, or is this something we want for the Kit API?

rohanKanojia: It was added here to ensure that any `Namespace`/`Project` is applied before any resource. It was there in the Mojo already; I just moved logic to a method. I think you're right. It should be called from `applyEntities()` method.

rohanKanojia: Anyway, whatever it was supposed to do seems broken. I've created https://github.com/eclipse/jkube/issues/511 for `Namespace` fragment creation issues. We might need to modify `DefaultNamespaceEnricher` a bit.

manusa (approved):
LGTM

Issue #472, opened 2020-11-03
Refactor tests in `KeyStoreUtilTest` to avoid duplication


rohanKanojia:
### Description

We have these three tests in `KeyStoreUtilTest` which seem to be doing the same task but with different input values.

https://github.com/eclipse/jkube/blob/90439cb98fef038e8c991825d2a90ce414787fb6/jkube-kit/build/service/docker/src/test/java/org/eclipse/jkube/kit/build/service/docker/access/KeyStoreUtilTest.java#L46-L63

We should move this common logic into some a helper method and make tests call that method instead.

deepak-gittyup:
Hi Rohan,
I am a newbie to open-source and would like to start with this project  
Can I pick this up ?

rohanKanojia:
sure, feel free to pick this up.

deepak-gittyup:
Hi,
I think I made the necessary changes to address this task and opened a PR request: https://github.com/eclipse/jkube/pull/476
It looks like, it failed 2 checks and I don't quite understand what to do next here
Can you help me out over here ?

rohanKanojia:
@deepak-gittyup : You need to create an Eclipse Account and sign Eclipse Contributor Agreement[0] in order to contribute to any Eclipse Foundation project

[0] https://www.eclipse.org/legal/ECA.php

PR #479, opened 2020-11-03
Fix: Refactor tests in `KeyStoreUtilTest` to avoid duplication (#472)

deepak-gittyup:
## Description
Fixes: #472 

## Type of change

 - [x] Chore (non-breaking change which doesn't affect codebase;

## Checklist
 - [ ] ~~I Added [CHANGELOG](../CHANGELOG.md) entry~~
 - [ ] ~~I have updated the [documentation](../kubernetes-maven-plugin/doc) accordingly~~
 - [ ] ~~I tested my code in Kubernetes~~
 - [ ] ~~I tested my code in OpenShift~~

### DIFF
--- a/jkube-kit/build/service/docker/src/test/java/org/eclipse/jkube/kit/build/service/docker/access/KeyStoreUtilTest.java
+++ b/jkube-kit/build/service/docker/src/test/java/org/eclipse/jkube/kit/build/service/docker/access/KeyStoreUtilTest.java
@@ -13,13 +13,15 @@
  */
 package org.eclipse.jkube.kit.build.service.docker.access;
 
-import java.security.GeneralSecurityException;
-import java.security.KeyStore;
-import java.security.PrivateKey;
-
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
+
+import java.io.FileNotFoundException;
+import java.net.URL;
+import java.security.GeneralSecurityException;
+import java.security.KeyStore;
+import java.security.PrivateKey;
 
 import static org.junit.Assert.assertNotNull;
 
@@ -36,7 +38,7 @@ public class KeyStoreUtilTest {
     public void createKeyStore() throws Exception {
         KeyStore keyStore = KeyStoreUtil.createDockerKeyStore(getFile("certpath"));
         KeyStore.PrivateKeyEntry pkEntry = (KeyStore.PrivateKeyEntry) keyStore.getEntry("docker",
-                                                                                        new KeyStore.PasswordProtection("docker".toCharArray()));
+                new KeyStore.PasswordProtection("docker".toCharArray()));
         assertNotNull(pkEntry);
         assertNotNull(pkEntry.getCertificate());
         assertNotNull(keyStore.getCertificate("cn=ca-test,o=internet widgits pty ltd,st=some-state,c=cr"));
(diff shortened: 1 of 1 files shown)


dengliming:
I suggest using `junit parameterized tests `

deepak-gittyup:
> I suggest using `junit parameterized tests `

Hi @dengliming 
From what I can see in the code, all the tests don't seem to have the same body
* There are 5 tests and 3 of those (`loadPrivateKeyPKCS8`, `loadPrivateKeyECDSA`, `loadPrivateKeyDefault`) fall in one category having same body with the only difference being the private key file. For this case, it makes sense to use parameterized tests
* But the next 2 tests (`loadInvalidPrivateKey` and `createKeyStore`), although they need an input private key path/cert path, I see that they have a different body and perform slightly different assertions from the other 3

From what I understood, to support parameterized tests, I need to separate the 5 tests into 2 separate classes: 3 + 2
Have I misunderstood anything here ?

deepak-gittyup:
Hi @rohanKanojia 
In the previous few PR attempts, there were 2 check failures and the reasons were:
* My git config was pointing to a different user.email other than the one that I had used to register in the eclipse community 
* I had passed incorrect key to one of the tests which had caused the coverage to drop by 0.01%
All other checks had passed
After correcting both of them, I was expecting all checks to pass. Yet 2 of them failed and from what I can see, they don't seem to be related to any of my change (Apologies if I am wrong here)

Is this a one time unexpected failure ? 
If so how do I rerun the checks again ? 
Should I do a dummy commit to trigger it ?

rohanKanojia:
Yes, it seems unrelated to your changes. I've restarted the workflow :+1:

deepak-gittyup:
> Yes, it seems unrelated to your changes. I've restarted the workflow 👍

Thanks @rohanKanojia 
One check still seems to have failed
From what I understand, looks like the shutdown hook in `PortForwardService` seems to have not been called while running `PortForwardServiceTest`
Perhaps some kind of unexpected/abrupt termination ?

Is it okay to go ahead despite this ?

rohanKanojia:
Yes, I think it's fine. Need to check with @manusa if we should split tests into two classes for parametrized approach  as suggested by @dengliming   or is it fine this way

manusa:
Hi,

Issue #472 was probably unnecessary since these tests were already quite readable and semantic.

However, since the issue is here, the only valid approach is via parameterized tests.
The current solution makes the tests more opaque since the test logic is hidden in a method with no semantic meaning `loadAndTestPrivateKey` (It tells what it does, but neither how nor what it asserts).

So if you want to keep at it, create the parameterized tests (find an example [here](https://github.com/manusa/jkube/blob/e03e87cca97bf703f9a8a6e2605b00dafb48b87d/jkube-kit/jkube-kit-micronaut/src/test/java/org/eclipse/jkube/micronaut/MicronautUtilsGetMicronautConfigurationTest.java#L31))

Yes, for clarity you should split tests. Each parameterized suite should go on its own class. The name of the class should be declarative enough to suggest what we are testing.

rohanKanojia:
@deepak-gittyup: Hi, polite ping. Are you still working on this patch?

alex81sp:
I can pick it up

alex81sp:
This PR could be closed,
completed here https://github.com/eclipse/jkube/pull/531

rohanKanojia:
Closing this since the issue is fixed by #531

PR #498, opened 2020-11-17
Update Kubernetes Client to v4.13.0

### LINKED ISSUE #387: Ingress generations fails, TLS issue?
### Description
I cannot get Ingress support working with both the XML method and with the Resource Fragment option

If I use the following fragment (which works on my k8s instance):

```yaml
apiVersion: networking.k8s.io/v1
metadata:
  name: my-ingress
  annotations:
    ingress.kubernetes.io/rewrite-target: /
    cert-manager.io/cluster-issuer: "letsencrypt-prod"
spec:
  tls:
  - hosts:
    - my.host.com
    secretName: letsencrypt-prod
  rules:
  - host: my.host.com
    http:
      paths:
        - path: /
          pathType: Prefix
          backend:
            service:
              name: my-service
              port: 
                number: 8080
```
I get the following error:
`Execution default-cli of goal org.eclipse.jkube:kubernetes-maven-plugin:1.0.0:resource failed: No resource type found for:networking.k8s.io/v1#Ingress`
### Info

* Eclipse JKube version : 1.0.0
* Maven version (`mvn -v`) : 3.6.0

* Kubernetes / Red Hat OpenShift setup and version :
Kubernetes 1.19

add the above fragment in src/main/jkube/ingress.yaml and you'll get the error.

rohanKanojia:
Fix #387

+ Updated Kubernetes Client to v4.13.0
+ Added a test to verify `networking.k8s.io/v1` `Ingress` is loaded

CQs:
- [X] https://dev.eclipse.org/ipzilla/show_bug.cgi?id=22822
- [X] https://dev.eclipse.org/ipzilla/show_bug.cgi?id=22823
## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ Usage:
 * Fix #268: Generator and HealthCheck enrichers for Micronaut framework (JVM)
 * Fix #488: Controller enricher replica count can be set to `0` when ResourceConfig is provided
 * Fix #485: Filter with placeholders in Dockerfile is broken 
+* Fix #387: Update Fabric8 Kubernetes Client to v4.13.0 to support `networking.k8s.io/v1` `Ingress`
 
 ### 1.0.2 (2020-10-30)
 * Fix #429: Added quickstart for Micronaut framework
--- a/jkube-kit/common/src/main/java/org/eclipse/jkube/kit/common/util/KubernetesHelper.java
+++ b/jkube-kit/common/src/main/java/org/eclipse/jkube/kit/common/util/KubernetesHelper.java
@@ -78,7 +78,6 @@ import io.fabric8.kubernetes.client.ConfigBuilder;
 import io.fabric8.kubernetes.client.KubernetesClient;
 import io.fabric8.kubernetes.client.KubernetesClientException;
 import io.fabric8.kubernetes.client.Watch;
-import io.fabric8.kubernetes.client.Watcher;
 import io.fabric8.kubernetes.client.dsl.FilterWatchListDeletable;
 import io.fabric8.kubernetes.client.dsl.LogWatch;
 import io.fabric8.kubernetes.client.dsl.NonNamespaceOperation;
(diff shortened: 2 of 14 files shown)


manusa:
In case we generate an Ingress, what will the default version be?

rohanKanojia:
Right now `IngressEnricher` seems to be generating `extensions/v1beta1` `Ingress`. `Ingress` generated using this enricher would be of apiVersion `extensions/v1beta1`

rohanKanojia:
I'm not sure whether the default `Ingress` generated by `IngressEnricher` should be `networking.k8s.io/v1`. This was introduced in Kubernetes 1.19.0. If we change it from `extensions/v1beta1`to `networking.k8s.io/v1`, I'm afraid it might break backward compatibility.

jkube-kit/config/service/src/main/java/org/eclipse/jkube/kit/config/service/kubernetes/KubernetesUndeployService.java
```diff
@@ -108,7 +108,7 @@ private void undeployCustomResources(String namespace, File resourceDir, Resourc
   // Visible for testing
   String currentNamespace(List<HasMetadata> entities) {
     for (HasMetadata entity : entities) {
-      if (entity instanceof Namespace || entity instanceof Project) {
+      if ((entity instanceof Namespace || entity instanceof Project) && entity.getMetadata() != null) {
```
manusa, line 111: When did you get a NullPointerException for this case?

manusa (approved):

jkube-kit/config/service/src/main/java/org/eclipse/jkube/kit/config/service/kubernetes/KubernetesUndeployService.java
rohanKanojia: When I remove this I was getting NullPointerException due to this:
https://github.com/eclipse/jkube/blob/d2234119936e8ab95c258f4538f57deefb5fa075/jkube-kit/config/service/src/test/java/org/eclipse/jkube/kit/config/service/kubernetes/KubernetesUndeployServiceTest.java#L93

manusa: Ok, but that's because the Test is providing an invalid Project/Namespace.
In a real-world scenario, I understand that if this happens, we shouldn't ignore it. Unless I'm wrong, receiving a Namespace/Project lacking Metadata should be a flag of something going very wrong so maybe reporting as an Exception is something we want.
With the current approach, we're simply assuming there's no namespace.

PR #534, opened 2021-01-12
Fix #515: Properties now gets resolved in CustomResource fragments

### LINKED ISSUE #515: When deploying CRDs to kubernetes properties aren't replaced in deployment yaml files
### Info
* Eclipse JKube version : 1.0.2
* Maven version (`mvn -v`) :
```
Apache Maven 3.6.3
Maven home: /usr/share/maven
Java version: 11.0.9, vendor: Oracle Corporation, runtime: /usr/lib/jvm/java-11-oracle
OS name: "linux", version: "5.4.0-56-generic", arch: "amd64", family: "unix"

```
* Kubernetes / Red Hat OpenShift setup and version : minikube latest version ( kubernetes version 1.19.4 ) and EKS cluster kubernetes v1.17.4

* If it's a bug, how to reproduce : Use my sampleKerberosProject, first apply test_resource.yaml custom resource definition, then mvn k8s:resource && mvn k8s:apply
You will get error, which will be solved if you in src/main/jkube/test-cr.yml replace ${app_namespace} with any namespace.

I was looking into your code, and quick fix is to replace resourceDir with OutputDirectory. But probably better solution is to replace properties in getCustomResourcesFileToNameMap function. 

* Sample Reproducer Project : https://github.com/marespane/sampleKerberosProject

rohanKanojia:
Fix #515 

CustomResource fragments are now deserialized/serialized into GenericCustomResource so that they are processed with standard kubernetes types

## Type of change

## Checklist


manusa (changes requested):
Check https://github.com/eclipse/jkube/issues/515#issuecomment-766984743

jkube-kit/common/src/main/java/org/eclipse/jkube/kit/common/util/ResourceUtil.java
```diff
@@ -50,6 +55,42 @@ public static boolean jsonEquals(JsonObject first, JsonObject second) {
...
+            kubernetesResources.add(resource);
+        }
+        return kubernetesResources;
+    }
+
+    private static KubernetesResource getHasMetadataOrGenericResource(ResourceFileType resourceFileType, Map<String, Object> item) {
+        try {
+            return convertValue(resourceFileType, item, KubernetesResource.class);
+        } catch (IllegalArgumentException illegalArgumentException) {
+            return convertValue(resourceFileType, item, GenericCustomResource.class);
```
manusa, line 86: When is this going to happen?
`GenericCustomResource` implements `KubernetesResource`, I don't understand what we're covering here.

manusa: OK, so after running the tests, the question is just the opposite, why is this method not just (or even removed)?:
```
    private static GenericCustomResource getHasMetadataOrGenericResource(ResourceFileType resourceFileType, Map<String, Object> item) {
        return convertValue(resourceFileType, item, GenericCustomResource.class);
    }
```

jkube-kit/common/src/main/java/org/eclipse/jkube/kit/common/util/KubernetesHelper.java
```diff
@@ -509,7 +513,7 @@ public static void handleKubernetesClientException(KubernetesClientException e,
     }
 
     public static Set<HasMetadata> loadResources(File manifest) throws IOException {
-        Object dto = ResourceUtil.load(manifest, KubernetesResource.class);
+        Object dto = ResourceUtil.loadKubernetesResourceList(manifest);
```
manusa, line 516: This will break the if-else cause in line 521 dto will never be an instance of `Template`

manusa: I'm tracking this down to be only used on the undeploy phase goal

jkube-kit/common/src/main/java/org/eclipse/jkube/kit/common/util/ResourceUtil.java
manusa: OK, I see the point now, that when deserializing to `KubernetesResource.class` we will make use of the KubernetesDeserializer, as opposed to what happens when using the `GenericCustomResource.class` as target, since this one overrides the JsonDeserialize annotation.

However, this is really hidden and the code doesn't really help to understand what's going on.

jkube-kit/common/src/main/java/org/eclipse/jkube/kit/common/util/KubernetesHelper.java
manusa: https://github.com/eclipse/jkube/pull/534/files#r568642847

Method names are really confusing here.

manusa:
Please check: https://github.com/rohanKanojia/jkube/pull/15

jkube-kit/config/service/src/main/java/org/eclipse/jkube/kit/config/service/kubernetes/KubernetesUndeployService.java
```diff
@@ -114,14 +112,28 @@ String currentNamespace(List<HasMetadata> entities) {
...
+    }
+  }
 
-    Map<String, Object> customResource = unmarshalCustomResourceFile(customResourceFile);
-    Map<String, Object> objectMeta = (Map<String, Object>)customResource.get("metadata");
-    String name = objectMeta.get("name").toString();
-    logger.info("Deleting Custom Resource %s %s", KubernetesHelper.getFullyQualifiedApiGroupWithKind(crdContext), name);
-    KubernetesClientUtil.doDeleteCustomResource(jKubeServiceHub.getClient(), crdContext, namespace, name);
+  private void deleteCustomResourceIfPresent(GenericCustomResource customResource, String namespace, CustomResourceDefinitionContext crdContext) {
+    Map<String, Object> cr = KubernetesClientUtil.doGetCustomResource(jKubeServiceHub.getClient(), crdContext, namespace, customResource.getMetadata().getName());
```
manusa, line 122: Why are we checking for presence before deleting?

manusa: @rohanKanojia:
```
It is due to behavior in raw customresource API that if resource doesn't exist. It fails with 404
in standard DSL we return a boolean value on delete(), if 404 is received we return false.
I guess we should align raw customresource delete API with this
``` 
@manusa
```
OK, that is bad then
Yes, we need to fix this in the client.
```

manusa (approved):

Issue #602, opened 2021-03-04
oc:resource do not generate route


rafaeltuelho:
### Description

`oc:resource` do not generate openshift route

### Info
even explicitly configuring the `openshift-maven-plugin` with:
```xml
...
            <configuration>
              <verbose>false</verbose>
              <enricher>
                <config>
                  <jkube-openshift-route>
                    <generateRoute>true</generateRoute>
                    <tlsInsecureEdgeTerminationPolicy>Allow</tlsInsecureEdgeTerminationPolicy>
                    <tlsTermination>edge</tlsTermination>
                  </jkube-openshift-route>
                </config>
              </enricher> 
...
```

* Eclipse JKube version : 1.1.1
* Maven version (`mvn -v`) :
```
Apache Maven 3.6.3 (cecedd343002696d0abb50b32b541b8a6ba2883f)
Maven home: /usr/local/Cellar/maven/3.6.3_1/libexec
Java version: 11.0.2, vendor: Oracle Corporation, runtime: /Library/Java/JavaVirtualMachines/openjdk-11.0.2.jdk/Contents/Home
Default locale: en_US, platform encoding: UTF-8
OS name: "mac os x", version: "10.15.6", arch: "x86_64", family: "mac"
```
* Kubernetes / Red Hat OpenShift setup and version : 
```
Client Version: openshift-clients-4.6.0-202006250705.p0-166-g299b6af53
Server Version: 4.6.3
Kubernetes Version: v1.19.0+9f84db3
```

* If it's a bug, how to reproduce :
```
mvn clean oc:resource -Popenshift
```
* If it's a feature request, what is your use case :

* Sample Reproducer Project : *[https://github.com/rafaeltuelho/my-business-automation-showcase/tree/master/business-application-service]*

manusa:
_**WIP: I'm trying to reproduce the issue.**_

When I run `mvn clean oc:resource -Popenshift` with the project as is in https://github.com/rafaeltuelho/my-business-automation-showcase/blob/0fa29a20c99adbde48e456a00589ff29b7736ccc/business-application-service/pom.xml a route gets successfully created. :heavy_check_mark: 

If I remove the [route.yml](https://github.com/rafaeltuelho/my-business-automation-showcase/blob/0fa29a20c99adbde48e456a00589ff29b7736ccc/business-application-service/src/main/jkube/route.yml) fragment, the route is no longer created :heavy_multiplication_x: 


[...]

PR #616, opened 2021-03-17
Make expose labels created by ExposeEnricher persistent

rnetuka:
~~According to docs https://www.eclipse.org/jkube/docs/openshift-maven-plugin#resource-route-generation, Openshift services should be exposed by default. Only when you add expose=false label, no routes will be generated.~~

ExposeEnricher should propagate expose: true label to RouteEnricher for services exposed by default (running on port 80 etc.) Routes should be generated for such services.

Actual behavior:
~~If no expose label is provided, the check _isExposedService(serviceMetadata)_ returns false and no route is created.~~
ExposeEnricher doesn't propagate expose: true label to RouteEnricher. The check _isExposedService(serviceMetadata)_ returns false and no route is created.

Fix proposal:
~~If no expose label is provided, add it by default to expose=true~~
Propagate expose: true label created by ExposeEnricher to RouteEnricher

Related to https://issues.redhat.com/browse/ENTESB-15988

### DIFF
--- a/jkube-kit/enricher/generic/src/main/java/org/eclipse/jkube/enricher/generic/openshift/ExposeEnricher.java
+++ b/jkube-kit/enricher/generic/src/main/java/org/eclipse/jkube/enricher/generic/openshift/ExposeEnricher.java
@@ -13,22 +13,23 @@
  */
 package org.eclipse.jkube.enricher.generic.openshift;
 
-import io.fabric8.kubernetes.api.model.HasMetadata;
+import io.fabric8.kubernetes.api.builder.TypedVisitor;
 import io.fabric8.kubernetes.api.model.KubernetesListBuilder;
 import io.fabric8.kubernetes.api.model.ObjectMeta;
-import io.fabric8.kubernetes.api.model.Service;
+import io.fabric8.kubernetes.api.model.ServiceBuilder;
 import io.fabric8.kubernetes.api.model.ServicePort;
 import io.fabric8.kubernetes.api.model.ServiceSpec;
+import org.eclipse.jkube.kit.common.util.KubernetesHelper;
 import org.eclipse.jkube.kit.config.resource.PlatformMode;
 import org.eclipse.jkube.kit.enricher.api.BaseEnricher;
 import org.eclipse.jkube.kit.enricher.api.JKubeEnricherContext;
-import org.eclipse.jkube.kit.common.util.KubernetesHelper;
 
 import java.util.Arrays;
 import java.util.HashSet;
 import java.util.List;
-import java.util.Map;
 import java.util.Set;
+
+import static org.eclipse.jkube.kit.enricher.api.util.KubernetesResourceUtil.containsLabelInMetadata;
 
 
 /**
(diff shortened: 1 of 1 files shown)


manusa:
Relates to #602

manusa:
Please, check this comment (https://github.com/eclipse/jkube/issues/602#issuecomment-791371801) and see how it aligns with the current proposed fix.

rnetuka:
@manusa I've checked your comment and I can confirm that ExposeEnricher works ok and adds expose: true label (since the service port is 80 -> exposed by default as you've said). 

However, **RouteEnricher** still fails. The expose: true label is not detected by _isExposedService(serviceMetadata)_ check (it returns _false_, so the route creation is skipped). Expected behavior: route is created.

<jkube.enricher.jkube-service.expose>true</jkube.enricher.jkube-service.expose> doesn't solve my problem since it addresses functionality that already works in my case. Providing _expose: true_ manually in _service.yml_ works, but I see it more like a workaround, since the route should be created by default in my case.

You can check yourself:
https://github.com/fabric8-quickstarts/spring-boot-cxf-jaxws, branch fuse-7.x.sb2.redhat
> mvn oc:deploy -Popenshift

manusa:
Hi @rnetuka 
Routes should be automatically created for applications with services that have ports exposed in the following set (80, 443, 8080, 9080, 9090, 9443), without any additional configuration.
From your changes, apparently, we'll now be exposing services in any port, which I'm not sure is the expected behavior (docs need to be updated too).
I'll thoroughly check the provided examples and logs, because I think I may be missing something, and get back to you.

rnetuka:
@manusa This is the problem, that routes are _not_ created in my case (fabric8-quickstarts). In **RouteService**, there is a check **isExposedService(serviceMetadata) ** and it checks for some labels in metadata. These labels are not present, so the route is not created.
https://github.com/eclipse/jkube/blob/master/jkube-kit/enricher/generic/src/main/java/org/eclipse/jkube/enricher/generic/openshift/RouteEnricher.java#L107
https://github.com/eclipse/jkube/blob/master/jkube-kit/enricher/api/src/main/java/org/eclipse/jkube/kit/enricher/api/util/KubernetesResourceUtil.java#L930

manusa:
Yes, I understand that. The question is why are those labels not in the generated Service.
If it's because it's not one of those "exposable" service ports (80, 443, 8080, 9080, 9090, 9443), or for some other reason.
Anyway, I'm starting to check now.

rnetuka:
@manusa The services from fabric8-quickstarts should be on port 80. When I was checking, the ExposeEnricher worked fine and generated expose: true label. However, in RouteEnricher, there is a ServiceBuilder (instead of service) and that is not modified by the ExposeEnricher. I think it needs to be mofidied too. 

See the updated PR to check what I mean.

manusa:
Checking ;)

manusa:
There seems to be something very wrong in the ExposeEnricher, I don't think that the labels get persisted, not sure how the expose label ends up in the Services right now though.

manusa:
I'm not really sure why the `ExposeEnricher` works right now, probably depending on the Fabric8 Kubernetes Client version things might or might not work as expected.

The fact is that `ExposeEnricher` should be using the visitor pattern instead of modifying built resources (which in reality should provide immutability to the Builder based items).

So basically what you need to do is remove the other part and keep yours:
- Move your new `TypedVisitor` to the `create` method
- Change the argument of the `enrichService` method to `ServiceBuilder`
- Make your visitor call the enrichService and refactor the logic there
  - Instead of using `editMetadata` use `editOrNewMetadata`

Also, you could use the time to make the webPorts field static (`WEB_PORTS`) ;)

If you want us to create an additional commit on top of yours with the suggested changes, just ask.

Thanks!

manusa (changes requested):
Requested changes are in previous comment

rnetuka:
Thanks for the review! I've made the changes to the PR.

manusa (approved):
LGTM, thx!

rohanKanojia (approved):

PR #629, opened 2021-03-26
Fix #624: Unable to override Image Name in Simple Dockerfile Mode with `jkube.generator.name`

### LINKED ISSUE #624: Not able to override Image Name in Simple Dockerfile Mode with `jkube.generator.name`
When using simple Dockerfile mode, we're not able to override the default image name using `jkube.generator.name` property. Seems like we always assume image name to be group:artifact:latest in zero dockerfile mode and don't consider whether `jkube.generator.name` property is set or not, we do pass a defaultImageName but somehow it seems to be passed as null:

https://github.com/eclipse/jkube/blob/fd9003cf04fd2bcd7407e87ffc30dbf41d8a24b7/jkube-kit/build/api/src/main/java/org/eclipse/jkube/kit/build/api/helper/DockerFileUtil.java#L249-L252

In order to reproduce this, you can go to quickstarts/maven/docker-file-simple and try overriding image name by adding some property. I noticed this when testing @ghassen1khalil sample: https://github.com/ghassen1khalil/jkube-error-reproducer

rohanKanojia:
Fix #624 
Modify Mojos logic to consider image name configured from properties

## Type of change

## Checklist

### DIFF
--- a/jkube-kit/common/src/main/java/org/eclipse/jkube/kit/common/util/PropertiesUtil.java
+++ b/jkube-kit/common/src/main/java/org/eclipse/jkube/kit/common/util/PropertiesUtil.java
@@ -13,6 +13,8 @@
  */
 package org.eclipse.jkube.kit.common.util;
 
+import io.fabric8.kubernetes.client.utils.Utils;
+
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URL;
@@ -38,5 +40,24 @@ public class PropertiesUtil {
       }
     }
     return ret;
+  }
+
+  /**
+   * Return first Non Null set property from a set of provided properties
+   *
+   * @param properties {@link Properties} in a given project
+   * @param keys an array of property key values to find
+   * @return a string which is first non null value found for the provided list
+   */
+  public static String getValueFromProperties(Properties properties, String... keys) {
+    for (String property : keys) {
+      if (properties.containsKey(property)) {
+        String value = properties.get(property).toString();
+        if (Utils.isNotNullOrEmpty(value)) {
+          return value;
+        }
+      }
+    }
+    return null;
   }
 }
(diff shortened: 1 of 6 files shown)


rohanKanojia:
I think I removed usage of properties in DockerFileUtil. Now properties needed for defaultImageName are resolved in Mojos and resolvedImageName is passed to DockerFileUtil

jkube-kit/common/src/main/java/org/eclipse/jkube/kit/common/util/PropertiesUtil.java
```diff
@@ -39,4 +39,20 @@ public static Properties getPropertiesFromResource(URL resource) {
...
   }
+
+  /**
+   * Return first Non Null set property from a set of provided properties
+   *
+   * @param properties {@link Properties} in a given project
+   * @param keys an array of property key values to find
+   * @return a string which is first non null value found for the provided list
+   */
+  public static String getValueFromProperties(Properties properties, String[] keys) {
```
manusa, line 50: Changing to:
```suggestion
  public static String getValueFromProperties(Properties properties, String... keys) {
```
will make life easier for method users and invocations will look clearer.

jkube-kit/common/src/main/java/org/eclipse/jkube/kit/common/util/PropertiesUtil.java
```diff
@@ -39,4 +39,20 @@ public static Properties getPropertiesFromResource(URL resource) {
...
+    for (String property : keys) {
+      if (properties.containsKey(property)) {
+        return properties.get(property).toString();
```
manusa, line 53: It might make sense to check if the property is not blank either

(get property, check if not null, check if not blank, return)

manusa (changes requested):
see comment

manusa (approved):

manusa (approved):

Issue #632, opened 2021-03-26
Add support for Quarkus Fast Jar Packaging


rohanKanojia:
[Quarkus 1.12](https://quarkus.io/blog/quarkus-1-12-0-final-released/) was released last month which made Fast-jar packaging as default. Earlier we used to run `*-runner.jar` in previous versions, but now it's replaced by this:
```
java -jar target/quarkus-app/quarkus-run.jar
```
You can read more about it [here](https://github.com/quarkusio/quarkus/wiki/Migration-Guide-1.12). 

Right now users can't use jkube in zero-configuration  mode with Fast Jar. JKube fails with not being able to find `-runner.jar`:
```
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.eclipse.jkube:kubernetes-maven-plugin:1.1.1:build (default-cli) on project rosa-fruits-app: Execution default-cli of goal org.eclipse.jkube:kubernetes-maven-plugin:1.1.1:build failed: Can't find single file with suffix '-runner.jar' in /home/rohaan/work/repos/rosa-fruits-app/target (zero or more than one files found ending with '-runner.jar') ->
```

 We would need to make changes to QuarkusGenerator to handle Fast-jar packaging as well. We just need to build ImageConfiguration like done in this [Dockerfile here](https://github.com/kameshsampath/rosa-fruits-app/blob/5dd38fd7bbe8ec05d76f175a56d1bb383f4062e8/src/main/docker/Dockerfile.jvm#L3)

### How to Reproduce:
- Try Cloning @kameshsampath 's repo https://github.com/kameshsampath/rosa-fruits-app 
- Remove Image XML configuration
- Try running `k8s:build` or `oc:build`

manusa:
You can easily work around this issue by setting `<quarkus.package.type>uber-jar</quarkus.package.type>` as stated in the generated README.md when bootstraping a Quarkus project >= 1.12

manusa:
Improvement is related to #425 when trying to optimize image builds.

From the generator perspective, we should identify possible assembly strategies (uber, native, fast-jar) and then create the assembly configuration based on that.

kameshsampath:
@rohanKanojia @manusa  ^^ PR #649

rohanKanojia:
@kameshsampath : Thanks a lot for the PR but there is one PR from @dloiacono to fix the same issue: https://github.com/eclipse/jkube/pull/638

dloiacono:
[...]

PR #638, opened 2021-03-30
Add support for Quarkus Fast Jar Packaging

dloiacono:
Signed-off-by: dloiacono <dloiacono@gmail.com>

## Description
Quarkus 1.12 was released last month which made Fast-jar packaging as default. Earlier we used to run *-runner.jar in previous versions, but now it's replaced by this:

java -jar target/quarkus-app/quarkus-run.jar
You can read more about it here.

Right now users can't use jkube in zero-configuration mode with Fast Jar. JKube fails with not being able to find -runner.jar:

[ERROR] Failed to execute goal org.eclipse.jkube:kubernetes-maven-plugin:1.1.1:build (default-cli) on project rosa-fruits-app: Execution default-cli of goal org.eclipse.jkube:kubernetes-maven-plugin:1.1.1:build failed: Can't find single file with suffix '-runner.jar' in /home/rohaan/work/repos/rosa-fruits-app/target (zero or more than one files found ending with '-runner.jar') ->
We would need to make changes to QuarkusGenerator to handle Fast-jar packaging as well. We just need to build ImageConfiguration like done in this Dockerfile here

## Type of change

## Checklist
 - [X] No new bugs, code smells, etc. in [SonarCloud](https://sonarcloud.io/dashboard?id=jkubeio_jkube) report

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ Usage:
 * Fix #602: Add documentation regarding automatic Route generation
 * Fix #630: Document usage for `jkube.build.switchToDeployment` flag
 * Fix #647: Resource configuration can now add annotations and labels to ServiceAccount
+* Fix #632: Add support for Quarkus Fast Jar Packaging
 
 ### 1.2.0 (2021-03-31)
 * Fix #529: `.maven-dockerignore`, `.maven-dockerexclude`, `.maven-dockerinclude` are no longer supported
--- a/jkube-kit/jkube-kit-quarkus/pom.xml
+++ b/jkube-kit/jkube-kit-quarkus/pom.xml
@@ -64,5 +64,6 @@
             <groupId>org.assertj</groupId>
             <artifactId>assertj-core</artifactId>
         </dependency>
+
     </dependencies>
 </project>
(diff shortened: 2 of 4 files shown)


manusa:
Intends to fix #632 
This is not really the expected approach because you can configure Quarkus to produce an uber-jar or not depending on your config (see `quarkus.package.type` property).

So  the assumption that depending on a given Quarkus version you will get one packaging or the other is not completely right.

We should basically try to detect the main JAR package with some other means. Also, taking into account other configurations such as `quarkus.package.runner-suffix` that might affect the generated JAR.

Since al of this is hard, what might work in most cases is detecting the target/*.jar that matches the application name as prefix. Also adding an enricher configuration in case this could be overridden manually.

dloiacono:
Sounds reasonable to me.
And what about the "quarkus-app" folder in the target?

manusa:
I'm not sure, I haven't had the chance to analyze the non uber-jar mode.

It seems that when running in fast-jar mode, `quarkus-app` is always generated (https://quarkus.io/guides/maven-tooling#using-fast-jar), right?

If the above is true, and stable across different versions (<1.12 too), we should try to infer the packaging mode by the structure of the target/build directory:
- Is native: proceed with native assembly
- Is fast-jar (---> `quarkus-app` exists): Generated assembly should copy files in `target/quarkus-app/*` to `/deployments`
- fast-jar not detected (---> `target/lib` subdirectory exists + fat jar in `target`): Generate assembly as we are doing now
- Other: Copy fattest jar in `/target` to `/deployments` (same as JavaExec)

Something along those lines should work.

dloiacono:
Seems that only when the `quarkus-app` forder is present in the target the logic have to be changed.

dloiacono:
@rohanKanojia I forgot to add also the Openshift version of the Helm Push mojo.
Is the case to add it?

rohanKanojia:
@dloiacono : I think it makes sense to add it, @manusa  WDYT?. You can add OpenShiftHelmPushMojo which will just extend original one just like most of it's goals

manusa (approved):

Issue #488, opened 2020-11-12
jkube.enricher.jkube-controller.replicaCount cannot be set to 0


manusa:
### Description

When using the `jkube.enricher.jkube-controller.replicaCount` or enricher XML configuration element, 0 is not supported.

In order to allow 0 as a value on the property, the following XML configuration is required:
```xml
<plugin>
 <groupId>org.eclipse.jkube</groupId>
 <artifactId>openshift-maven-plugin</artifactId>
 <version>${jkube.version}</version>
 <configuration>
   <resources>
     <replicas>0</replicas>
   </resources>
 </configuration>
</plugin>
```

The problem seems to be this default value:
https://github.com/eclipse/jkube/blob/e11392a64b1f63ae64c9082e6a4dd9dad2bf420d/jkube-kit/config/resource/src/main/java/org/eclipse/jkube/kit/config/resource/ResourceConfig.java#L74

### Info

* Eclipse JKube version: 1.0.2

manusa:
https://github.com/eclipse/jkube/blob/aa8d431bca1ad8031138fbc80b88dc960d75f230/jkube-kit/enricher/api/src/main/java/org/eclipse/jkube/kit/enricher/api/BaseEnricher.java#L177

rohanKanojia:
Hi, I'm not able to reproduce this issue. I tested on v1.0.2 and this seems to work as expected for me. I've tested it on my sample:
https://github.com/rohanKanojia/eclipse-jkube-demo-project . It just has basic zero configuration option like this:
```xml
            <plugin>
                <groupId>org.eclipse.jkube</groupId>
                <artifactId>openshift-maven-plugin</artifactId>
                <version>${jkube.version}</version>
            </plugin>
```

Here is what I did.

1. Deployed application using standard goals `mvn oc:build oc:resource oc:apply`. Application gets deployed:
```
jkube : $ oc get pods -w
NAME                           READY     STATUS      RESTARTS   AGE
random-generator-1-4d64q       1/1       Running     0          58s
random-generator-s2i-1-build   0/1       Completed   0          1m
```
2. Issue Stop Equivalent: `mvn oc:resource oc:apply -Djkube.enricher.jkube-controller.replicaCount=0` . Application Pod gets terminated:
```
jkube : $ oc get pods -w
NAME                           READY     STATUS      RESTARTS   AGE
random-generator-1-4d64q       1/1       Running     0          58s
random-generator-s2i-1-build   0/1       Completed   0          1m
[...]

PR #640, opened 2021-03-30
Re-introduce jkube.replicas property

rnetuka:
**Enhancement**

Current state:
When user wants to specify replica count, he has to use either _-Djkube.enricher.jkube-controller.replicaCount_ or _-Djkube.enricher.jkube-controller-from-configuration.replicaCount_ based on if resource fragments are used or not. This can be very confusing for users.

Proposal:
Unify both approaches (DefaultControllerEnricher and ControllerViaPluginConfigurationEnricher) to use _-Djkube.replicaCount_ instead

Reference issue: https://issues.redhat.com/browse/ENTESB-16080

### DIFF
--- a/kubernetes-maven-plugin/plugin/src/main/java/org/eclipse/jkube/maven/plugin/mojo/build/ResourceMojo.java
+++ b/kubernetes-maven-plugin/plugin/src/main/java/org/eclipse/jkube/maven/plugin/mojo/build/ResourceMojo.java
@@ -196,6 +196,9 @@ public class ResourceMojo extends AbstractJKubeMojo {
     @Parameter(property="jkube.interpolateTemplateParameters", defaultValue = "true")
     private Boolean interpolateTemplateParameters;
 
+    @Parameter(property = "jkube.replicas")
+    private Integer replicas;
+
     @Component
     private MavenProjectHelper projectHelper;
 
@@ -251,6 +254,9 @@ public class ResourceMojo extends AbstractJKubeMojo {
         realResourceDir = ResourceUtil.getFinalResourceDir(resourceDir, environment);
         if (namespace != null && !namespace.isEmpty()) {
             resources = ResourceConfig.toBuilder(resources).namespace(namespace).build();
+        }
+        if (replicas != null) {
+            resources = ResourceConfig.toBuilder(resources).replicas(replicas).build();
         }
         final ResourceServiceConfig resourceServiceConfig = ResourceServiceConfig.builder()
             .project(javaProject)
--- a/kubernetes-maven-plugin/doc/src/main/asciidoc/inc/goals/build/_jkube-resource.adoc
+++ b/kubernetes-maven-plugin/doc/src/main/asciidoc/inc/goals/build/_jkube-resource.adoc
@@ -445,4 +445,8 @@ endif::[]
 | Domain added to the Service ID when creating Kubernetes Ingresses or OpenShift routes.
 | `jkube.domain`
 
+| *replicas*
+| Number of replicas for the container.
+| `jkube.replicas`
+
 |===


manusa:
The [JIRA issue](https://issues.redhat.com/browse/ENTESB-16080) makes a valid point in the following statement:

> Previously with fabric8 solution adopted was:
>
> `mvn fabric8:start -Dfabric8.replicas=2`

This field is still present in Eclipse JKube, and is the preferred way to set the number of replicas.

This is how it's used via XML configuration elements:
https://github.com/eclipse/jkube/blob/cf88735616f98ca48b03c756ba5741bf0c92ce8c/quickstarts/maven/xml-config/pom.xml#L210

Apparently we might have lost the way to configure that field using properties (was a field part of the [StartMojo](https://github.com/fabric8io/fabric8-maven-plugin/blob/f85754379dbef6ae9d7c3a366da6ce1499665229/plugin/src/main/java/io/fabric8/maven/plugin/mojo/develop/StartMojo.java#L39), removed in 4.x versions of FMP).

This is indirectly related to #488 and its fix #499 (for completeness).

Your solution addresses the problem. However, we're actually trying to move away from reading properties in implementation logic within the jkube-kit module, since this is hard to address when JKube is used as a standard Java API (or in the future when we implement the Gradle plugins).

The proposed solution would be to recover the `jkube.replicas` field so that it can be used to initialize the [ResourceConfig#replicas](https://github.com/eclipse/jkube/blob/efe3e902ea9ef571be5c32839cfa64321908aa18/jkube-kit/config/resource/src/main/java/org/eclipse/jkube/kit/config/resource/ResourceConfig.java#L73) field.

So (if I'm not mistaken) the fix should be as easy as adding the following code in [ResourceMojo](https://github.com/eclipse/jkube/blob/a80ecbdb2bba680b173b32e6104133b112e6041c/kubernetes-maven-plugin/plugin/src/main/java/org/eclipse/jkube/maven/plugin/mojo/build/ResourceMojo.java).
```java
@Parameter(property = "jkube.replicas")
private Integer replicas;
```
and modify the `initJKubeServiceHubBuilder` method implementation:
```java
        if (namespace != null && !namespace.isEmpty()) {
            resources = ResourceConfig.toBuilder(resources).namespace(namespace).build();
        }
        if (replicas != null) {
            resources = ResourceConfig.toBuilder(resources).replicas(replicas).build();
        }
```

We need to make sure that if the XML configuration is provided, it's not overwritten by the above implementation. And documentation in the [appropriate section](https://www.eclipse.org/jkube/docs/kubernetes-maven-plugin#Supported-Properties-Resource).

manusa (changes requested):
See previous comment

rnetuka:
@manusa do we make the changes _on top of_ -Djkube.replicaCount, or _instead of them_?

manusa:
> @manusa do we make the changes on top of the current -Djkube.replicaCount change, or instead of it?

I'd prefer instead of.

This way we only need to modify the Mojo (+docs) and the Enrichers remain untouched. I tried my suggested changes, and at first glance should solve the issue.

rnetuka:
@manusa It is strange, but if I use 

`@Parameter(property = "jkube.replicas")
private Integer replicas;`

the value is not injected. Other existing parameters work fine (for example _namespace_), but newly added don't. I was wondering if I have to register the property somewhere.

Example of running:
`> mvn oc:resource -Djkube.replicas=5 -Djkube.namespace=rnetuka`

rohanKanojia:
Are you adding this replicas mojo parameter to ResourceConfig builder here?: https://github.com/eclipse/jkube/blob/master/kubernetes-maven-plugin/plugin/src/main/java/org/eclipse/jkube/maven/plugin/mojo/build/ResourceMojo.java#L253

manusa:
> ~Are you adding this replicas mojo parameter to ResourceConfig builder here?: https://github.com/eclipse/jkube/blob/master/kubernetes-maven-plugin/plugin/src/main/java/org/eclipse/jkube/maven/plugin/mojo/build/ResourceMojo.java#L253~
> 
> I think we would need to check how this parameter is getting resolved here: https://github.com/eclipse/jkube/blob/master/jkube-kit/enricher/api/src/main/java/org/eclipse/jkube/kit/enricher/api/BaseEnricher.java#L178

It gets resolved the same way the configuration element.

***

I basically did the changes this morning and was working fine. I'll retry and paste some screenshots.

manusa:
This is the result for an execution:
```
quickstarts/maven/spring-boot $ mvnDebug -Djkube.version=1.2.0-SNAPSHOT -Popenshift clean oc:resource -Djkube.replicas=0
```

image

image

image

manusa:
If you're testing with any of the quickstarts, make sure to test the right version `-Djkube.version=1.2.0-SNAPSHOT` (I miss this all the time)

rnetuka:
I was finally able to test the changes. I had to add a check if the value hasn't been provided in pom.xml. Without it, it would be overwritten.

I suggest to do a reshape of ResourceConfig in the future and provide methods like putIfAbsent(int replicas). 

Please, see the new PR.

kubernetes-maven-plugin/plugin/src/main/java/org/eclipse/jkube/maven/plugin/mojo/build/ResourceMojo.java
```diff
@@ -252,6 +255,10 @@ protected RuntimeMode getRuntimeMode() {
         if (namespace != null && !namespace.isEmpty()) {
             resources = ResourceConfig.toBuilder(resources).namespace(namespace).build();
         }
+        // only overwrite replica count if it hasn't been provided in pom.xml
+        if (replicas != null && ResourceConfig.toBuilder(resources).build().getReplicas() == null) {
```
manusa, line 259: I think one of my [comments](https://github.com/eclipse/jkube/pull/640#issuecomment-810753726) might be misleading and lead you to add the &&... check.

That **shouldn't** be there, because if that's users wouldn't be able to override the value via command line.

```suggestion
        if (replicas != null) {
```

_My comment was more related to the fact of declaring the `replicas` field as an Integer (as opposed to an int) and nullchecking the value (replicas !=null). If the field was to be declared as a primitive type, it would be confusion and difficult to know when the parameter should override the value or not._

Sorry :pray:

manusa (approved):
LGTM, thx!

manusa (approved):

rohanKanojia (approved):

rnetuka:
One final update. I've removed the comment above the condition check since the check for xml value was also removed.

Issue #658, opened 2021-04-09
Apply configuration in DependencyEnricher


rnetuka:
In DependencyEnricher, configuration is not taken into account. For example if you have two projects, one depends
on another, and you run

```
 > cd project_A
 > mvn oc:resource -Djkube.replicas=2
```

it creates
_target/classes/META-INF/jkube/openshift/project_A_deploymentconfig.yml_
with `spec.replicas: 2`
_target/classes/META-INF/jkube/openshift/project_B_deploymentconfig.yml_
with `spec.replicas: 1`

It would be nice to overwrite the replica count in DependencyEnricher just as DefaultControllerEnricher does. Also, apply other configuration that makes sense.

manusa:
Hi @rnetuka ,
Thx for reporting the issue.
As discussed, let us further investigate the issue. DependencyEnricher hasn't been revisited in a long time and it's possible that we're missing other configurations too (besides replicas).

rnetuka:
PR https://github.com/eclipse/jkube/pull/660

manusa:
I think the solution for this issue might be as simple as switching the order of the Enrichers
These lines:
https://github.com/eclipse/jkube/blob/4809716a08ab90ddf9ec7e7d329e5b7e020630a9/openshift-maven-plugin/plugin/src/main/resources/META-INF/jkube/profiles-default.yml#L32
https://github.com/eclipse/jkube/blob/4809716a08ab90ddf9ec7e7d329e5b7e020630a9/kubernetes-maven-plugin/plugin/src/main/resources/META-INF/jkube/profiles-default.yml#L32

Should be moved to the top.

Have you tried that?

manusa:
> Have you tried that?

I've tried this with a few more changes like changing the override from `enrich` to `create`, but the problem is really not with the `DependencyEnricher` implementation (maybe the order should be fixed either way). 

The problem is about what we expect from this Enricher and the behavior of the `-Djkube.replicas` modifier.

The enricher basically merges **materialized** YAML configurations that come from dependencies. The fact that they are materialized, means that these YAMLs are to be considered as fragments of the dependent project.

So following with the -Djkube.replicas modifier, the behavior of this flag is to set the number of replicas for **generated** manifests. This means that for the case where a manifest is completely generated from JKube (**and** didn't contain a `spec.replicas` field), the value of the replicas will be set for us.
[...]

PR #660, opened 2021-04-09
Feature #658: Apply configuration in DependencyEnricher

rnetuka:
Solution for #658 

Current solution is for replica count only. If you have any tips what to add, just let me know.


rohanKanojia:
Thanks a lot for your PR :+1: . Could you please add a test to verify that provided fix works as per our expectations?

rnetuka:
Amendments:
1. Test case added
2. /jkube-kit/enricher/generic/src/test/resources/jenkins-4.0.41.jar!/META-INF contains **fabric8** directory instead of **jkube**. I've added a fix for that (I hope it works this way since GitHub won't show me the changes...)
3. Inside the new function `item` could be any of `Deployment`, `DeploymentConfig`, `DeploymentBuilder` and `DeploymentConfigBuilder`. The _if-else_ plus _cast_ is extremely ugly, so any tips how to make it better are welcome

rnetuka:
Replaced by https://github.com/eclipse/jkube/pull/675

PR #675, opened 2021-04-20
Feature #658: Apply configuration in DependencyEnricher

rnetuka:
Issue: https://github.com/eclipse/jkube/issues/658

I tried to get this feature as simple as possible. It basically reverted the -Djkube.replicas property provided earlier on ResourceMojo and created ReplicaCountEnricher instead. The task of this enricher is to ovewrite the replica count in Deployment/DeploymentConfig spec.

I didn't use `enforcedReplicas`, nor the `ResourceConfig` itself because I think it could have lead to confusion with overwriting the value in different places and to potential bugs. This way, if you disable `ReplicaCountEnricher` (or don't use `-Djkube.replicas`), the value is left untouched.

Any thoughts and comments are welcome.


rohanKanojia:
Could you please try fixing license issues by running `mvn -N license:format`?

I think along with Deployment/DeploymentConfig we should also consider Kubernetes resources used in [DefaultControllerEnricher](https://github.com/eclipse/jkube/blob/master/jkube-kit/enricher/generic/src/main/java/org/eclipse/jkube/enricher/generic/DefaultControllerEnricher.java#L67) basically any resource which manages pods and has `spec.replicas` field

rnetuka:
@rohanKanojia Are we able to modify Kubernetes API model and add _HasSpec_ interface on such resources?

rohanKanojia:
Sorry for the late reply, I don't think we can modify kubernetes model. Why do you need it?

rnetuka:
@rohanKanojia  It would be great to cast all of the resources to one common interface while setting the replica count instead of repeating this construct on every possible resource 
`builder.accept(new TypedVisitor<Replicable>() {
     @Override
     public void visit(Replicable replicable) {
         deploymentBuilder.editOrNewSpec().withReplicas(replicas).endSpec();
     }
});`

rohanKanojia:
looks neat :+1: . Unfortunately, right now we don't have anything like this in Kubernetes Model. Our model gets generated from go code and is not persisted. 

We might make changes in go generator to make a specific set of resources implement an additional Replicable interface while generating schema: https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-model-generator/pkg/schemagen/generate.go#L227

We're already doing it for distinguishing between Namespaced and Cluster scoped resources. I think we might be able to do it for these Replicable resources too

manusa:
Sorry for the late review.
I'm really not convinced about reading a property from within the Enricher, since this is not that straightforward when using JKube Kit as an API.
Anyway, since we need to release, I'll approve and merge. We can see later on how to better address this problem (it's the same for other properties such as `jkube.createExternalUrls`, `jkube.domain`, etc.)

manusa (approved):

PR #694, opened 2021-05-10
Fix #671: ApplyMojo not considering namespace configured from XML configuration

### LINKED ISSUE #671: ApplyMojo not considering namespace configured from XML configuration
### Info
When I'm trying to do `k8s:apply` from a project where namespace is configured from XML rather than `jkube.namespace` property. I'm getting this error:
```
[INFO] k8s: Creating a ServiceAccount from kubernetes.yml namespace default name rokumar-serviceaccount
[ERROR] k8s: Failed to create ServiceAccount from kubernetes.yml. io.fabric8.kubernetes.client.KubernetesClientException: Namespace mismatch. Item namespace:rokumar. Operation namespace:default.. ServiceAccount(apiVersion=v1, automountServiceAccountToken=null, imagePullSecrets=[], kind=ServiceAccount, metadata=ObjectMeta(annotations=null, clusterName=null, creationTimestamp=null, deletionGracePeriodSeconds=null, deletionTimestamp=null, finalizers=[], generateName=null, generation=null, labels={app=jkube-xml-resourceconfig, provider=jkube, version=1.0.0-SNAPSHOT, group=org.eclipse.jkube}, managedFields=[], name=rokumar-serviceaccount, namespace=rokumar, ownerReferences=[], resourceVersion=null, selfLink=null, uid=null, additionalProperties={}), secrets=[], additionalProperties={}): io.fabric8.kubernetes.client.KubernetesClientException: Namespace mismatch. Item namespace:rokumar. Operation namespace:default.
[INFO] BUILD FAILURE
```
I think we're not considering namespace configured from ResourceConfig here:

https://github.com/eclipse/jkube/blob/3af6339ddfa16efbe42ef6e71687f583d74856ee/kubernetes-maven-plugin/plugin/src/main/java/org/eclipse/jkube/maven/plugin/mojo/build/ApplyMojo.java#L235-L239
* Eclipse JKube version : v1.2.0
* Maven version (`mvn -v`) :
```
Apache Maven 3.6.3 (cecedd343002696d0abb50b32b541b8a6ba2883f)
Maven home: /opt/apache-maven-3.6.3
Java version: 11.0.9, vendor: Oracle Corporation, runtime: /usr/lib/java/jdk-11.0.9
OS name: "linux", version: "5.11.14-200.fc33.x86_64", arch: "amd64", family: "unix"

```
* Kubernetes / Red Hat OpenShift setup and version :

  - Clone the reproducer project
  - create a namespace `rokumar` or change namespace configured in pom to your desired namespace
  - `mvn k8s:resource k8s:apply`

* Sample Reproducer Project : https://github.com/r0haaaan/jkube-xml-resourceconfig/tree/namespace-bug

rohanKanojia:
## Description
Fix #671 

Refactor ApplyService to consider namespace in resource metadata over the default namespace to enable applying resources in multiple namespaces
+ When no namespace is configured via XML or property, namespace gets picked from clusterAccess.getNamespace
+ When no namespace is configured via XML or property, but namespace is set in resource fragments, that particular namespace would be picked for those specific resources. Other resources would have clusterAccess.getNamespace namespace picked
+ When user provides namespace via `jkube.namespace` property. ALL resources are applied into namespace specified via `jkube.namespace`  property

## Type of change

## Checklist

### DIFF
--- a/jkube-kit/config/service/src/main/java/org/eclipse/jkube/kit/config/service/kubernetes/KubernetesUndeployService.java
+++ b/jkube-kit/config/service/src/main/java/org/eclipse/jkube/kit/config/service/kubernetes/KubernetesUndeployService.java
@@ -82,9 +82,10 @@ public class KubernetesUndeployService implements UndeployService {
 
   protected Consumer<HasMetadata> resourceDeleter(String namespace) {
     return resource -> {
-      logger.info("Deleting resource %s %s/%s", KubernetesHelper.getKind(resource), namespace, KubernetesHelper.getName(resource));
+      String undeployNamespace = resolveEffectiveNamespace(resource, namespace);
+      logger.info("Deleting resource %s %s/%s", KubernetesHelper.getKind(resource), undeployNamespace, KubernetesHelper.getName(resource));
       jKubeServiceHub.getClient().resource(resource)
-          .inNamespace(namespace)
+          .inNamespace(undeployNamespace)
           .withPropagationPolicy(DeletionPropagation.BACKGROUND)
           .delete();
     };
@@ -92,9 +93,10 @@ public class KubernetesUndeployService implements UndeployService {
 
   protected Consumer<HasMetadata> customResourceDeleter(String namespace) {
     return customResource -> {
+      String undeployNamespace = resolveEffectiveNamespace(customResource, namespace);
       GenericCustomResource genericCustomResource = (GenericCustomResource) customResource;
       CustomResourceDefinitionList crdList = jKubeServiceHub.getClient().apiextensions().v1beta1().customResourceDefinitions().list();
-      deleteCustomResourceIfCustomResourceDefinitionContextPresent(genericCustomResource, namespace, getCrdContext(crdList, genericCustomResource));
+      deleteCustomResourceIfCustomResourceDefinitionContextPresent(genericCustomResource, undeployNamespace, getCrdContext(crdList, genericCustomResource));
     };
   }
 
@@ -120,6 +122,13 @@ public class KubernetesUndeployService implements UndeployService {
     } catch (Exception exception) {
       logger.error("Unable to undeploy %s %s/%s", apiVersionAndKind, namespace, name);
     }
+  }
+
+  private String resolveEffectiveNamespace(HasMetadata entity, String fallbackNamespace) {
+    if (KubernetesHelper.getNamespace(entity) != null) {
+      return KubernetesHelper.getNamespace(entity);
+    }
+    return fallbackNamespace;
   }
 
   protected JKubeServiceHub getjKubeServiceHub() {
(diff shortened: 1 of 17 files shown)


jkube-kit/config/service/src/main/java/org/eclipse/jkube/kit/config/service/kubernetes/KubernetesUndeployService.java
```diff
@@ -82,19 +82,21 @@ private void undeployResources(String namespace, List<HasMetadata> entities) {
 
   protected Consumer<HasMetadata> resourceDeleter(String namespace) {
     return resource -> {
-      logger.info("Deleting resource %s %s/%s", KubernetesHelper.getKind(resource), namespace, KubernetesHelper.getName(resource));
+      String undeployNamespace = resolveEffectiveNamespace(resource, namespace);
```
manusa, line 85: Why isn't the namespace resolved after line 68 and passed on? (Just like we're doing now)

rohanKanojia: This is done to handle the case of multiple namespaces. In case resources have different namespaces in their metadata, it's given more precedence over standard namespace.

rohanKanojia:
I have a doubt regarding this case:
> Non-generated or generated resources have namespace configured in fragment (this namespace may be different to that configured via XML ---> I have additional fragments/resources to deploy that don't belong to my application's domain). This specific, granular, per-resource configuration should take precedence over the more global XML configuration.

Once namespace is configured in XML like this:
```xml
<resources>
  <namespace>rokumar-stage</namespace>
</resources>
```
It will modify all resources' `metadata.namespace` field. How would our generated resources end up having different namespaces in this case? 

Shall we refactor DefaultNamespaceEnricher to only add namespace if not already present? We can add another flag (ex `jkube.namespace.forced`)  to override this behavior 
https://github.com/eclipse/jkube/blob/297b84661950cf8e1cab173919743967f02f1b55/jkube-kit/enricher/generic/src/main/java/org/eclipse/jkube/enricher/generic/DefaultNamespaceEnricher.java#L108

jkube-kit/config/service/src/test/java/org/eclipse/jkube/kit/config/service/ApplyServiceTest.java
```diff
@@ -362,6 +367,72 @@ public void testGetK8sListWithNamespaceFirstWithProject() {
...
+
+        // Then
+        collector.assertEventsRecordedInOrder("configmap-ns1-create", "serviceaccount-default-create", "ingress-ns2-create");
+        assertEquals(6, mockServer.getMockServer().getRequestCount());
+        applyService.setFallbackNamespace(null);
+        applyService.setNamespace(configuredNamespace);
+    }
+
+    @Test
+    public void testApplyToMultipleNamespacesOverriddenWhenNamespaceConfigured() throws InterruptedException {
```
manusa, line 406: I'm not sure what this test does compared to `testApplyToMultipleNamespaceNoNamespaceConfigured`

manusa (approved):
LGTM

Issue #705, opened 2021-05-25
JIB image build in Windows fails


manusa:
### Description
When running a JIB image build in Windows it fails with:
```
[ERROR] oc: Failed to execute the build [Error when building JIB image]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  6.289 s
[INFO] Finished at: 2021-05-25T12:12:12+02:00
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.eclipse.jkube:openshift-maven-plugin:1.3.0:build (default) on project spring-boot-helm: Failed to execute the build: Error when building JIB image: C:\Users\mnurisan\Downloads\delete\jkube\quickstarts\maven\spring-boot-helm\target\docker\spring-boot-helm\1.3.0\build\deployments\spring-boot-helm-1.3.0.jar: The process cannot access the file because it is being used by another process. -> [Help 1]
```

WIP

### Info

* Eclipse JKube version : 1.3.0

* If it's a bug, how to reproduce :
```shell
$ cd quickstarts\maven\spring-boot-helm
$ mvn clean package -Popenshift -Djkube.build.strategy=jib
```

manusa:
Relates to #300

Issue #300, opened 2020-07-20
Refactor AssemblyManager so that AssemblyFileEntries can be computed/shared easilly


manusa:
## Description

AssemblyManager should provide an easy way to retrieve the list of `AssemblyFileEntry` that will be included in the assembly.

Required by #299

PR #706, opened 2021-05-25
fix: JIB assembly works on Windows

manusa:
## Description
Fix: #705 

## Type of change

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ Usage:
 ./scripts/extract-changelog-for-version.sh 1.3.37 5
 ```
 ### 1.4.0-SNAPSHOT
+* Fix #705: JIB assembly works on Windows
 
 ### 1.3.0
 * Fix #497: Assembly descriptor removed but still in documentation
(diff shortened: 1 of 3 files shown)


rohanKanojia:
JIB build model is working fine on my machine

rohanKanojia:
Umm, sorry could you please provide some context? I see you moved `getAssemblyTarArchive` call below . How does this fix this issue?

manusa:
> Umm, sorry could you please provide some context? I see you moved `getAssemblyTarArchive` call below . How does this fix this issue?

Yes, sorry.

For the JIB assembly process we do something quite ugly since we need to reuse part of what is done for the Docker assembly process and those methods weren't designed with reusability in mind (hence the relation of #705 with #300).

The problem is that we are copying twice the files to the destination temporary directory and in Windows, when the Jar artifact is "replaced" there is an error since for some reason it gets loaded in memory. The current approach is just a quick fix, the proper fix should be implemented in scope of #300. We already did some progress in that area, and probably the double copy approach call should no longer be necessary. However, there are still some procedures run in the `getAssemblyTarArchive` method which aren't externalized, so maybe it's not a good idea to completely remove/replace that call yet.

rohanKanojia (approved):

PR #719, opened 2021-06-11
Fix #672: Add missing fields in ProbeConfig for configuring Readiness/Liveness Probes

### LINKED ISSUE #672: Improve our Liveness/Readiness Probe XML configuration support
Right now our support for [Liveness/Readiness probes](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#http-probes) is okay but I think it's not up to date with latest releases of kubernetes. 

I see some things missing:
- Support for configuring custom headers `httpHeaders`
- Configuring `periodSeconds`
- ~Configuring `terminationGracePeriodSeconds`~ (preview feature) (see https://github.com/eclipse/jkube/issues/720)

rohanKanojia:
## Description
Fix #672 
+ Add missing fields in ProbeConfig for configuring liveness/readiness
+ Added a check to keep host null in case it matches localhost

## Type of change

## Checklist

### DIFF
--- a/jkube-kit/enricher/api/src/main/java/org/eclipse/jkube/kit/enricher/handler/ProbeHandler.java
+++ b/jkube-kit/enricher/api/src/main/java/org/eclipse/jkube/kit/enricher/handler/ProbeHandler.java
@@ -15,6 +15,7 @@ package org.eclipse.jkube.kit.enricher.handler;
 
 import io.fabric8.kubernetes.api.model.ExecAction;
 import io.fabric8.kubernetes.api.model.HTTPGetAction;
+import io.fabric8.kubernetes.api.model.HTTPHeader;
 import io.fabric8.kubernetes.api.model.IntOrString;
 import io.fabric8.kubernetes.api.model.Probe;
 import io.fabric8.kubernetes.api.model.TCPSocketAction;
@@ -24,8 +25,12 @@ import org.eclipse.jkube.kit.config.resource.ProbeConfig;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.List;
+import java.util.Map;
+import java.util.function.BiConsumer;
+import java.util.function.Supplier;
 
 import static org.apache.commons.lang3.StringUtils.isNotBlank;
+import static org.eclipse.jkube.kit.common.util.KubernetesHelper.convertMapToHTTPHeaderList;
 
 
 /**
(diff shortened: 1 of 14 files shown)


manusa:
Please rebase, please fix the code smell

rohanKanojia:
I'm not using IntSupplier for a reason here. It doesn't handle nullable integers well

manusa:
> I'm not using IntSupplier for a reason here. It doesn't handle nullable integers well

Then remove the warning and provide a hint as to why we use Supplier instead:
```java
    @SuppressWarnings("java:S4276") // IntSupplier throws NullPointerException when unboxing null Integers
    private void setTimeoutInProbeIfNotNull(Probe probe, Supplier<Integer> integerSupplier, BiConsumer<Integer, Probe> probeConsumer) {
```

jkube-kit/enricher/api/src/main/java/org/eclipse/jkube/kit/enricher/handler/ProbeHandler.java
```diff
@@ -93,6 +89,22 @@ private HTTPGetAction getHTTPGetAction(String getUrl) {
         }
     }
 
+    private String resolveHost(String host) {
+        if (host.equalsIgnoreCase("localhost")) {
+            // No need to set host in case of localhost HTTP request, will default to Pod IP
+            return null;
+        }
+        return host;
+    }
```
manusa, line 98: Why is this necessary?

rohanKanojia: This is done in order to support local probes. If we set the host to null, it is assumed to be pod's IP. This way when a user provides `<getUrl>http://localhost:8080/actuator/health</getUrl>`, it generates a probe like this:
```yaml
        readinessProbe:
          httpGet:
            path: /actuator/health
            port: 8080
            scheme: HTTP
```

manusa: I see what it does, but local probes were supported before (weren't they?), so why is this necessary now?

rohanKanojia: In order to provide local probes user needed to provide host omitted getUrl like this: `<getUrl>http://:8080/actuator/health</getUrl>` . I thought providing localhost could be a cleaner alternative

manusa: We should revert this (it won't consider cases such as 127.0.0.1) and it is misleading for users.
The host part of the [probe config](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#http-probes) should be discouraged
> host: Host name to connect to, defaults to the pod IP. You probably want to set "Host" in httpHeaders instead.

kubernetes-maven-plugin/doc/src/main/asciidoc/inc/goals/build/_jkube-resource.adoc
```diff
@@ -219,7 +219,17 @@ Probe configuration is used for configuring https://kubernetes.io/docs/tasks/con
 | Command to execute for probing.
 
 | `getUrl`
-| Probe this URL.
+| Probe URL for HTTP Probe. Configures HTTP probe fields like `host`, `scheme`, `path` etc by parsing URL. For example, a `<getUrl>http://:8080/health</getUrl>` would result in probe generated with fields set like this:
+
+  host: ""
+
+  path: /actuator/health
```
manusa, line 226: Shouldn't this be just `/health` ?

jkube-kit/enricher/api/src/main/java/org/eclipse/jkube/kit/enricher/handler/ProbeHandler.java
manusa: @rohanKanojia , please update the PR description to match the new behavior:
> Added a check to keep host null in case it matches localhost

PR #727, opened 2021-06-16
Fix #548: Define property for skipping cluster autodetect/offline mode

### LINKED ISSUE #548: Define property for skipping cluster autodetect mode and go straight to the Kubernetes offline mode
It would be nice if we could tell JKube to work offline on resources and push goals at least, 
so it skips the cluster autodetect mode, which can fail with ugly stacktraces if the  environment has a cluster but semiconfigured (for instant the aws credentials), and go straight to Kubernetes mode like the regular fallback does.
This prevents any unwanted interaction with the local cluster
  

Thanks mates!

rohanKanojia:
## Description
    
+ Add a property `jkube.offline` which which jkube would skip
   cluster auto detection and go straight to offline mode.

## Type of change

## Checklist

### DIFF
--- a/jkube-kit/config/service/src/main/java/org/eclipse/jkube/kit/config/service/JKubeServiceHub.java
+++ b/jkube-kit/config/service/src/main/java/org/eclipse/jkube/kit/config/service/JKubeServiceHub.java
@@ -68,13 +68,14 @@ public class JKubeServiceHub implements Closeable {
     private LazyBuilder<UndeployService> undeployService;
     private LazyBuilder<MigrateService> migrateService;
     private LazyBuilder<DebugService> debugService;
+    private final boolean offline;
 
     @Builder
     public JKubeServiceHub(
             ClusterAccess clusterAccess, RuntimeMode platformMode, KitLogger log,
             ServiceHub dockerServiceHub, JKubeConfiguration configuration,
             BuildServiceConfig buildServiceConfig,
-            LazyBuilder<ResourceService> resourceService) {
+            LazyBuilder<ResourceService> resourceService, boolean offline) {
         this.clusterAccess = clusterAccess;
         this.platformMode = platformMode;
         this.log = log;
@@ -82,6 +83,7 @@ public class JKubeServiceHub implements Closeable {
         this.configuration = configuration;
         this.buildServiceConfig = buildServiceConfig;
         this.resourceService = resourceService;
+        this.offline = offline;
         init();
     }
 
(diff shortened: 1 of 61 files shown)


jkube-kit/config/service/src/main/java/org/eclipse/jkube/kit/config/service/JKubeServiceHub.java
```diff
@@ -68,62 +68,33 @@
     private LazyBuilder<UndeployService> undeployService;
     private LazyBuilder<MigrateService> migrateService;
     private LazyBuilder<DebugService> debugService;
+    private final boolean connectToCluster;
```
manusa, line 71: Is there any specific reason to call the configuration offline and then the Service Hub variable `connectToCluster`?

If the offline configuration is only used for this purpose and the `connectToCluster` variable is only set by the offline configuration, it would make much more sense to name them the same. It would also help in the future when searching for configurations and debugging.

jkube-kit/config/service/src/main/java/org/eclipse/jkube/kit/config/service/JKubeServiceHub.java
```diff
@@ -168,4 +139,57 @@ public DebugService getDebugService() {
         return debugService.get();
     }
 
+    private void initClusterDependentServiceLazyBuilders() {
```
manusa, line 142: Not all of these services seem to be dependent on cluster.
```suggestion
    private void initClusterAccessAndLazyBuilders() {
```

jkube-kit/config/service/src/test/java/org/eclipse/jkube/kit/config/service/JKubeServiceHubTest.java
```diff
@@ -59,10 +61,15 @@
   private BuildServiceConfig buildServiceConfig;
 
   private JKubeServiceHub.JKubeServiceHubBuilder commonInit() {
+    return commonInit(true);
+  }
+
+  private JKubeServiceHub.JKubeServiceHubBuilder commonInit(boolean connectToCluster) {
```
manusa, line 67: Follow the existing approach and take advantage of the builder.

If you were to add parameters for every override, then tests would be less declarative, fluent, and readable.

e.g. 
In line 118 you are overrding the cluster access **and** the platform mode.
---> `commonInit().platformMode(RuntimeMode.KUBERNETES).offline(true)`

manusa (approved):
LGTM, thx!

PR #748, opened 2021-06-30
Fix #253: JKubeServiceHub BuildService implementations are now loaded via ServiceLoader

rohanKanojia:
## Description
Fix #253 
+ Use ServiceLoader to load all Build Services and pick up service
  which is applicable in current context. This requires BuildService
  implementations to have zero-arg constructors so I need to set the
  elements from setters instead.

+ Added isApplicable(), setJKubeServiceHub() methods in
  BuildService interface

Signed-off-by: Rohan Kumar <rohaan@redhat.com>

## Type of change

 - [ ] Bug fix (non-breaking change which fixes an issue)
 - [X] Feature (non-breaking change which adds functionality)
 - [ ] Breaking change (fix or feature that would cause existing functionality to change
 - [ ] Chore (non-breaking change which doesn't affect codebase;
   test, version modification, documentation, etc.)

## Checklist
 - [X] I have read the [contributing guidelines](https://www.eclipse.org/jkube/contributing)
 - [X] I signed-off my commit with a user that has signed the [Eclipse Contributor Agreement](https://www.eclipse.org/legal/ECA.php)
 - [X] I Added [CHANGELOG](../CHANGELOG.md) entry
 - [X] I have implemented unit tests to cover my changes
 - [X] I have updated the [documentation](../kubernetes-maven-plugin/doc) accordingly
 - [ ] No new bugs, code smells, etc. in [SonarCloud](https://sonarcloud.io/dashboard?id=jkubeio_jkube) report
 - [X] I tested my code in Kubernetes
 - [X] I tested my code in OpenShift

PR #772, opened 2021-07-08
Fix #253: JKubeServiceHub BuildService implementations are now loaded via PluginServiceFactory

### LINKED ISSUE #253: Refactor JKubeServiceHub's BuildService election mechanism
Right now we have a really ugly way of selecting a particular `BuildService` 
https://github.com/eclipse/jkube/blob/1c916f9e76566e8d2bc0721156742bd04bf5161a/jkube-kit/config/service/src/main/java/org/eclipse/jkube/kit/config/service/JKubeServiceHub.java#L100-L110

This approach is fine for now but as we plan to increase support for multiple builds(like JIB, Buildah, Podman etc). We need to refactor this and make it generic. We can probably implement it like we're doing in case of Generators:
https://github.com/eclipse/jkube/blob/1c916f9e76566e8d2bc0721156742bd04bf5161a/kubernetes-maven-plugin/plugin/src/main/java/org/eclipse/jkube/maven/plugin/generator/GeneratorManager.java#L52-L60

manusa:
## Description
Supersedes closes #748
Fixes #253 

Complete refactor of the BuildService(s) and load process in JKubeServiceHub

## Type of change

 - [x] Chore (non-breaking change which doesn't affect codebase;

## Checklist

### DIFF
--- a/jkube-kit/config/service/src/main/java/org/eclipse/jkube/kit/config/service/openshift/OpenShiftBuildServiceUtils.java
+++ b/jkube-kit/config/service/src/main/java/org/eclipse/jkube/kit/config/service/openshift/OpenShiftBuildServiceUtils.java
@@ -55,7 +55,7 @@ public class OpenShiftBuildServiceUtils {
 
   protected static File createBuildArchive(JKubeServiceHub jKubeServiceHub, ImageConfiguration imageConfig) throws JKubeServiceException {
     // Adding S2I artifacts such as environment variables in S2I mode
-    final ArchiverCustomizer customizer = createS2IArchiveCustomizer(jKubeServiceHub, imageConfig);
+    final ArchiverCustomizer customizer = createS2IArchiveCustomizer(jKubeServiceHub.getBuildServiceConfig(), imageConfig);
     try {
       return jKubeServiceHub.getDockerServiceHub().getArchiveService()
           .createDockerBuildArchive(imageConfig, jKubeServiceHub.getConfiguration(), customizer);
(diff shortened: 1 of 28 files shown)


jkube-kit/config/service/src/main/java/org/eclipse/jkube/kit/config/service/openshift/OpenShiftBuildServiceUtils.java
```diff
@@ -0,0 +1,214 @@
...
+    } catch (IOException e) {
+      throw new JKubeServiceException("Unable to create the build archive", e);
+    }
+  }
+
+  /**
+   * Returns the applicable name for the S2I Build resource considering the provided {@link ImageName} and
+   * {@link BuildServiceConfig}.
+   */
+  protected static String computeS2IBuildName(BuildServiceConfig config, ImageName imageName) {
```
rohanKanojia, line 71: Why did you choose methods in this class to be protected? I don't see this utility class getting inherited anywhere?

rohanKanojia (approved):

jkube-kit/config/service/src/main/java/org/eclipse/jkube/kit/config/service/openshift/OpenShiftBuildServiceUtils.java
manusa: Methods from OpenShiftBuildService were moved to this class one by one. So the reason is probably because in the initial iterations the method was consumed from OpenShiftBuildService and later on the calling methods were moved to the Utils class too.

I'll change this.

jkube-kit/config/service/src/main/java/org/eclipse/jkube/kit/config/service/openshift/OpenshiftBuildService.java
```diff
@@ -260,6 +225,18 @@ protected String updateOrCreateBuildConfig(BuildServiceConfig config, OpenShiftC
         }
     }
 
+    private void initClient() {
+        ClusterAccess clusterAccess = jKubeServiceHub.getClusterAccess();
+        if (clusterAccess == null) {
+            clusterAccess = new ClusterAccess(log,
+                ClusterConfiguration.from(System.getProperties(), jKubeServiceHub.getConfiguration().getProject().getProperties()).build());
+        }
+        client = clusterAccess.createDefaultClient();
```
rohanKanojia, line 234: assigning value of `clusterAccess.createDefaultClient()` which can either be `DefaultKubernetesClient` or `DefaultOpenShiftClient` seems to `OpenShiftClient client` seems to be causing https://github.com/eclipse/jkube/issues/815 

In v1.3.0, this code was in JKubeServiceHub and `client` was of type `KubernetesClient`

Issue #882, opened 2021-09-03
OpenShiftResourceTask: implement oc:resource for Dockerfile mode equivalent functionality for OpenShift Gradle Plugin


manusa:
## Description

Child of #31
Relates to #834

Implement resource generation for Dockerfile mode.

- Any configuration setting should be part of the OpenShiftExtension class (Tasks should have no Inputs)
- Any logic present in the current Mojos should be moved to JKube-kit

PR #915, opened 2021-09-15
followup #910: Correct DOCKER_IMAGE_USER property in OpenShiftResourceTask

rohanKanojia:
## Description
This property was wrongly named in ResourceMojo, for maven plugins
this got fixed in #910.

Fix property name in OpenShiftResourceTask as well

## Type of change

## Checklist

### DIFF
--- a/gradle-plugin/it/src/it/dockerfile-simple/expected/openshift.yml
+++ b/gradle-plugin/it/src/it/dockerfile-simple/expected/openshift.yml
@@ -58,9 +58,9 @@ items:
             valueFrom:
               fieldRef:
                 fieldPath: metadata.namespace
-          image: gradle/dockerfile-simple:latest
+          image: "@endsWith('dockerfile-simple:latest')@"
           imagePullPolicy: IfNotPresent
-          name: gradle-dockerfile-simple
+          name: "@endsWith('dockerfile-simple')@"
           ports:
           - containerPort: 9080
             name: glrpc
@@ -71,8 +71,6 @@ items:
     - type: ConfigChange
     - imageChangeParams:
         automatic: true
-        containerNames:
-        - gradle-dockerfile-simple
         from:
           kind: ImageStreamTag
           name: dockerfile-simple:latest
--- a/gradle-plugin/kubernetes/src/main/java/org/eclipse/jkube/gradle/plugin/task/AbstractJKubeTask.java
+++ b/gradle-plugin/kubernetes/src/main/java/org/eclipse/jkube/gradle/plugin/task/AbstractJKubeTask.java
@@ -75,12 +75,8 @@ public abstract class AbstractJKubeTask extends DefaultTask implements Kubernete
     jKubeServiceHub = initJKubeServiceHubBuilder().build();
     ImageConfigResolver imageConfigResolver = new ImageConfigResolver();
     try {
-      resolvedImages = initImageConfiguration(
-          kubernetesExtension.getApiVersion().getOrNull(),
-          getBuildTimestamp(null, null, javaProject.getBuildDirectory().getAbsolutePath(), DOCKER_BUILD_TIMESTAMP),
-          javaProject, kubernetesExtension.images, imageConfigResolver, kitLogger,
-          kubernetesExtension.getFilter().getOrNull(),
-          this::customizeConfig);
+      resolvedImages = resolveImages(imageConfigResolver);
+
       enricherManager = new DefaultEnricherManager(JKubeEnricherContext.builder()
         .project(javaProject)
           .processorConfig(ProfileUtil.blendProfileWithConfiguration(ProfileUtil.ENRICHER_CONFIG,
@@ -173,5 +169,14 @@ public abstract class AbstractJKubeTask extends DefaultTask implements Kubernete
     } catch (IOException e) {
       throw new IllegalArgumentException("Cannot extract generator config: " + e, e);
     }
+  }
+
+  protected List<ImageConfiguration> resolveImages(ImageConfigResolver imageConfigResolver) throws IOException {
+    return initImageConfiguration(
+      kubernetesExtension.getApiVersion().getOrNull(),
+      getBuildTimestamp(null, null, javaProject.getBuildDirectory().getAbsolutePath(), DOCKER_BUILD_TIMESTAMP),
+      javaProject, kubernetesExtension.images, imageConfigResolver, kitLogger,
+      kubernetesExtension.getFilter().getOrNull(),
+      this::customizeConfig);
   }
 }
--- a/gradle-plugin/openshift/src/main/java/org/eclipse/jkube/gradle/plugin/task/OpenShiftResourceTask.java
+++ b/gradle-plugin/openshift/src/main/java/org/eclipse/jkube/gradle/plugin/task/OpenShiftResourceTask.java
@@ -14,14 +14,18 @@
 package org.eclipse.jkube.gradle.plugin.task;
 
 import org.eclipse.jkube.gradle.plugin.OpenShiftExtension;
+import org.eclipse.jkube.kit.build.service.docker.config.handler.ImageConfigResolver;
+import org.eclipse.jkube.kit.config.image.ImageConfiguration;
 import org.eclipse.jkube.kit.config.resource.RuntimeMode;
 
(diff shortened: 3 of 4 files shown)


gradle-plugin/kubernetes/src/main/java/org/eclipse/jkube/gradle/plugin/task/AbstractJKubeTask.java
```diff
@@ -97,6 +98,11 @@ public final void runTask() {
     run();
   }
 
+  public void addEffectiveImageUserAndEffectivePlatformModeProperty() {
+    // Should be overridden by child classes in case properties need
+    // to be added before resolving images
+  }
+
```
manusa, line 105: 2 notes:
- Why are you making this public? if its meant to be overridden by child classes, make it protected.
- Why do we need an extra method (so specific) when child classes can simply override the `initImageConfiguration`?

gradle-plugin/it/src/it/dockerfile-simple/expected/openshift.yml
```diff
@@ -58,9 +58,9 @@ items:
             valueFrom:
               fieldRef:
                 fieldPath: metadata.namespace
-          image: gradle/dockerfile-simple:latest
+          image: "@endsWith('dockerfile-simple:latest')@"
```
manusa, line 61: What's the complete name? can we keep it specific or is it not stable enough?

gradle-plugin/it/src/it/dockerfile-simple/expected/openshift.yml
```diff
@@ -71,12 +71,10 @@ items:
     - type: ConfigChange
     - imageChangeParams:
         automatic: true
-        containerNames:
-        - gradle-dockerfile-simple
         from:
           kind: ImageStreamTag
           name: dockerfile-simple:latest
-          namespace: gradle
+          namespace: "@ignore@"
```
manusa, line 77: https://github.com/eclipse/jkube/pull/915/files#r709786504

gradle-plugin/openshift/src/main/java/org/eclipse/jkube/gradle/plugin/task/OpenShiftResourceTask.java
```diff
@@ -20,7 +20,7 @@
 import java.util.Properties;
 
 public class OpenShiftResourceTask extends KubernetesResourceTask {
-  private static final String DOCKER_IMAGE_USER = "docker.image.user";
+  private static final String DOCKER_IMAGE_USER = "jkube.image.user";
```
manusa, line 23: This property (both for Mojo and Task is used later). We should expose that constant and use it here (and in Mojo) instead.

gradle-plugin/it/src/it/dockerfile-simple/expected/openshift.yml
rohanKanojia: Complete name is resolved as namespace if I'm logged into an OpenShift cluster. For my case, it gets resolved to `rokumar-dev`

gradle-plugin/it/src/it/dockerfile-simple/expected/openshift.yml
rohanKanojia: This gets evaluated to current namespace if you're logged into OpenShift cluster. I didn't want test to fail in case you're logged into an openshift cluster. So I thought of ignoring this field.

gradle-plugin/it/src/it/dockerfile-simple/expected/openshift.yml
manusa: So probably something needs to be tuned in the build.gradle configuration so no real cluster is checked during the test.

manusa: https://github.com/eclipse/jkube/blob/72529d1aa599e5065452850c996a1e1fe777da46/gradle-plugin/it/src/it/dockerfile-simple/build.gradle#L15

openshift equivalent should fix it, right?

gradle-plugin/kubernetes/src/main/java/org/eclipse/jkube/gradle/plugin/task/AbstractJKubeTask.java
rohanKanojia: I wanted to set `jkube.image.user` property just before resolving final images. Earlier property was being set after resolving images.

I made this method `public` since I wanted to use it in OpenShiftResourceTask, with protected access it would only be available in KubernetesResourceTask

Right now in `runTask`, `resolvedImages` is initialized. I couldn't find a way to set properties before initializing resolvedImages. `runTask` is set final so it can't be overridden.

manusa: > Right now in runTask, resolvedImages is initialized. I couldn't find a way to set properties before initializing resolvedImages. runTask is set final so it can't be overridden.

OK, I thought `initImageConfiguration` was part of the AbstractJKubeTask, it's a static import (hard to see in the GH interface).

Anyway, `addEffectiveImageUserAndEffectivePlatformModeProperty` is **too** specific.

Maybe having a `resolveImages` method that can be overridden might be more interesting.
Other alternatives could be afterInit/lateInit, configuration, etc. methods that are located within different places in the runTask method. However, this might be early abstraction and become a burden.

> I made this method public since I wanted to use it in OpenShiftResourceTask, with protected access it would only be available in KubernetesResourceTask

That's a misunderstanding from your side, it would be this case if it was package-private (no access modifier). Protected methods can always be overridden in subclasses (unless declared final somewhere along the chain).

gradle-plugin/it/src/it/dockerfile-simple/expected/openshift.yml
rohanKanojia: I'm trying but it doesn't seem to work. Looks like `.kube/config` file is read and namespace is getting picked from there:

https://github.com/eclipse/jkube/blob/f6cd5409d0e8a0e4cb2fb44ea1bd359c5c8489e9/jkube-kit/common/src/main/java/org/eclipse/jkube/kit/common/util/KubernetesHelper.java#L433-L435

gradle-plugin/kubernetes/src/main/java/org/eclipse/jkube/gradle/plugin/task/AbstractJKubeTask.java
rohanKanojia: I've added the resolveImages method in AbstractJKubeTask and overriding this from OpenShiftResourceTask

gradle-plugin/it/src/it/dockerfile-simple/expected/openshift.yml
manusa: OK, we can try to solve this later.

gradle-plugin/kubernetes/src/main/java/org/eclipse/jkube/gradle/plugin/task/AbstractJKubeTask.java
```diff
@@ -75,12 +75,8 @@ public final void runTask() {
     jKubeServiceHub = initJKubeServiceHubBuilder().build();
     ImageConfigResolver imageConfigResolver = new ImageConfigResolver();
     try {
-      resolvedImages = initImageConfiguration(
-          kubernetesExtension.getApiVersion().getOrNull(),
-          getBuildTimestamp(null, null, javaProject.getBuildDirectory().getAbsolutePath(), DOCKER_BUILD_TIMESTAMP),
-          javaProject, kubernetesExtension.images, imageConfigResolver, kitLogger,
-          kubernetesExtension.getFilter().getOrNull(),
-          this::customizeConfig);
+      resolveImages(imageConfigResolver);
```
manusa, line 78: This is not reusable as it relies on changes of state within the method.
The idea was to have a resolveImages method (just like you did), but that returned the list of images, so
```suggestion
      resolvedImages = resolveImages(imageConfigResolver);
```

manusa (approved):

manusa:
Relates to #882

PR #944, opened 2021-09-28
test: Extension from properties initial implementation

manusa:
## Description
 - test: Extension from properties initial implementation
 - refactor: Optimized Gradle unit test structure

## Type of change

 - [ ] Bug fix (non-breaking change which fixes an issue)
 - [ ] Feature (non-breaking change which adds functionality)
 - [ ] Breaking change (fix or feature that would cause existing functionality to change
 - [ ] Chore (non-breaking change which doesn't affect codebase;
   test, version modification, documentation, etc.)

## Checklist
 - [ ] I have read the [contributing guidelines](https://www.eclipse.org/jkube/contributing)
 - [ ] I signed-off my commit with a user that has signed the [Eclipse Contributor Agreement](https://www.eclipse.org/legal/ECA.php)
 - [ ] I Added [CHANGELOG](../CHANGELOG.md) entry
 - [ ] I have implemented unit tests to cover my changes
 - [ ] I have updated the [documentation](../kubernetes-maven-plugin/doc) accordingly
 - [ ] No new bugs, code smells, etc. in [SonarCloud](https://sonarcloud.io/dashboard?id=jkubeio_jkube) report
 - [ ] I tested my code in Kubernetes
 - [ ] I tested my code in OpenShift

Issue #857, opened 2021-08-21
[EPIC] Migrate to Mockito mocking framework from JMockit in unit tests


rohanKanojia:
### Description

Right now we're using [JMockit](https://jmockit.github.io/) in our tests for mocking the behavior of classes. We're discussed this in past as well but never got any chance to create an issue for this. We should try to move to [Mockito](https://site.mockito.org/) which is a more modern mocking framework.

### Goals
- Remove JMockit dependency
  - JMockit expectations should be replaced by when() statements (Mockito mocks)
  - JMockit verifications should be replaced by verify() statements (Mockito verifications)
- Assertions should be replaced by AssertJ
- Tests should be evaluated and refactored if applicable

Since this can be a big task, I think we should try doing it step by step removing Jmockit from each module and using mockito instead. Here are the modules which use Jmockit:

- build/api https://github.com/eclipse/jkube/issues/1585
- build/service/docker https://github.com/eclipse/jkube/issues/1586
- build/service/jib https://github.com/eclipse/jkube/issues/1587
- common https://github.com/eclipse/jkube/issues/1545
- config/image https://github.com/eclipse/jkube/issues/1584
- config/service https://github.com/eclipse/jkube/issues/1588
- enricher/api https://github.com/eclipse/jkube/issues/1589
- enricher/generic https://github.com/eclipse/jkube/issues/1591
- enricher/specific https://github.com/eclipse/jkube/issues/1592
- generator/api https://github.com/eclipse/jkube/issues/1593
- generator/java-exec https://github.com/eclipse/jkube/issues/1594
- generator/karaf https://github.com/eclipse/jkube/issues/1595
- generator/webapp https://github.com/eclipse/jkube/issues/1596
- jkube-kit-micronaut https://github.com/eclipse/jkube/issues/1597
- jkube-kit-quarkus https://github.com/eclipse/jkube/issues/1598
- jkube-kit-spring-boot https://github.com/eclipse/jkube/issues/1599
- jkube-kit-thorntail-v2 https://github.com/eclipse/jkube/issues/1600
- jkube-kit-vertx https://github.com/eclipse/jkube/issues/1601
- jkube-kit-wildfly-jar https://github.com/eclipse/jkube/issues/1602
- resource/helm https://github.com/eclipse/jkube/issues/1603
- resource/service https://github.com/eclipse/jkube/issues/1604
- watcher/api https://github.com/eclipse/jkube/issues/1605
[...]

PR #957, opened 2021-09-30
GradleUtil.extractProperty should also extract System Properties

rohanKanojia:
## Description
While testing #944, I noticed that Gradle project.getProperties() don't
contain System Properties which are passed to JVM. So some property
passed like like won't work:
```
gradle -Djkube.offline=true ocApply
```

Modify GradleUtil.extractProperty to include System.getProperties() as
well while converting Gradle Project to JavaProject

## Type of change

## Checklist
 - [x] I tested my code in OpenShift

### DIFF
--- a/gradle-plugin/kubernetes/src/main/java/org/eclipse/jkube/gradle/plugin/GradleUtil.java
+++ b/gradle-plugin/kubernetes/src/main/java/org/eclipse/jkube/gradle/plugin/GradleUtil.java
@@ -22,6 +22,7 @@ import java.util.Objects;
 import java.util.Properties;
 import java.util.Set;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import org.eclipse.jkube.kit.common.Dependency;
 import org.eclipse.jkube.kit.common.JavaProject;
@@ -82,11 +83,12 @@ public class GradleUtil {
   }
 
   private static Properties extractProperties(Project gradleProject) {
-    return gradleProject.getProperties().entrySet().stream().filter(e -> Objects.nonNull(e.getValue()))
-        .reduce(new Properties(), (acc, e) -> {
-          acc.put(e.getKey(), e.getValue());
-          return acc;
-        }, (acc, e) -> acc);
+    return Stream.concat(gradleProject.getProperties().entrySet().stream(), System.getProperties().entrySet().stream())
+      .filter(e -> Objects.nonNull(e.getValue()))
+      .reduce(new Properties(), (acc, e) -> {
+        acc.put(e.getKey(), e.getValue());
+        return acc;
+      }, (acc, e) -> acc);
   }
 
   private static List<Dependency> extractDependencies(Project gradleProject) {
@@ -132,5 +134,4 @@ public class GradleUtil {
         .filter(File::exists)
         .findFirst().orElse(null);
   }
-
 }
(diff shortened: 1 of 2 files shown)


manusa:
In Gradle you really set build properties via project properties, command-line option `-P`, `--project-prop`.

However, I guess this won't hurt either, so :+1:

gradle-plugin/kubernetes/src/main/java/org/eclipse/jkube/gradle/plugin/GradleUtil.java
```diff
@@ -82,11 +82,17 @@ public static JavaProject convertGradleProject(Project gradleProject) {
...
+    Properties gradleProperties = gradleProject.getProperties().entrySet().stream().filter(e -> Objects.nonNull(e.getValue()))
+      .reduce(new Properties(), (acc, e) -> {
+        acc.put(e.getKey(), e.getValue());
+        return acc;
+      }, (acc, e) -> acc);
+    if (gradleProperties != null) {
+      properties.putAll(gradleProperties);
+    }
+    properties.putAll(System.getProperties());
+    return properties;
```
manusa, line 95: Can be simplified to:
```suggestion
    return Stream.concat(gradleProject.getProperties().entrySet().stream(), System.getProperties().entrySet().stream())
        .filter(e -> Objects.nonNull(e.getValue()))
        .reduce(new Properties(), (acc, e) -> {
          acc.put(e.getKey(), e.getValue());
          return acc;
        }, (acc, e) -> acc);
```

We might also want to think about property precedence or priority - Should system properties override those provided by Gradle project?

gradle-plugin/kubernetes/src/test/java/org/eclipse/jkube/gradle/plugin/GradleUtilTest.java
```diff
@@ -76,12 +76,28 @@ public void extractProperties_withComplexMap_shouldReturnValidProperties() {
     final JavaProject result = convertGradleProject(project);
     // Then
     assertThat(result.getProperties())
-        .hasSize(2)
+        .isNotEmpty()
```
manusa, line 79: You can use our SystemMock class to deal with this so we have a predictable size and values

gradle-plugin/kubernetes/src/test/java/org/eclipse/jkube/gradle/plugin/GradleUtilTest.java
```diff
@@ -76,12 +76,28 @@ public void extractProperties_withComplexMap_shouldReturnValidProperties() {
...
+    when(project.getProperties()).thenReturn(Collections.emptyMap());
+
+    // When
+    final JavaProject result = convertGradleProject(project);
+
+    // Then
+    assertThat(result.getProperties())
+      .hasFieldOrPropertyWithValue("foo.property", null)
+      .hasEntrySatisfying("foo.property", e -> assertThat(e).isEqualTo("somevalue"));
+    System.clearProperty("foo.property");
```
manusa, line 98: Use SystemMock to be able to provide a reliable test for this.

We can also add a test for property precedence.

gradle-plugin/kubernetes/src/test/java/org/eclipse/jkube/gradle/plugin/GradleUtilTest.java
rohanKanojia: SystemMock seems to be using JMockit for mocking:

https://github.com/eclipse/jkube/blob/6afbb4f371cec30e90b895490ccdd060ec53503b/jkube-kit/common-test/src/main/java/org/eclipse/jkube/kit/common/SystemMock.java#L19-L22

I'm not sure how to proceed since we already have plans to replace JMockit by Mockito in future. Shall I add a todo here for now to refactor this once https://github.com/eclipse/jkube/issues/857 is resolved?

manusa: Let's keep it this way. Once we merge this branch we can complete #958

manusa (approved):

PR #1046, opened 2021-10-20
Changed the sort method interface for easy readability

Simba-97:
Made the required changes. Changed Collections.sort(profiles, Collections.<Profile>reverseOrder()); in https://github.com/eclipse/jkube/blob/ee2390178dcad8a9cc3a9cb6d5e8c5d70a162a04/jkube-kit/profile/src/main/java/org/eclipse/jkube/kit/profile/ProfileUtil.java to    profiles.sort(Collections.reverseOrder()); for easy understanding.


Simba-97:
@rohanKanojia Hey rohan, I have removed the redundant comment as you advised. Thanks :)

PR #1140, opened 2021-11-23
Doc and examples: Avoid using prefix KUBERNETES_ on env vars

### LINKED ISSUE #1132: Env vars with KUBERNETES_ name prefix
In certain example fragments in the documentation it is shown that you can define your own env var and the name `KUBERNETES_NAMESPACE` is used in the example.

I believe the env var name prefix `KUBERNETES_` is being used by OpenShift ?

I think the documentation **should not encourage a home-made env var with this particular name prefix**. 

It would be better to show example like this:

```yaml
spec:
  ...
  template:
    spec:
      containers:
        - name: my-container
          env:
            - name: MY_POD_NAMESPACE
              valueFrom:
                fieldRef:
                  apiVersion: v1
                  fieldPath: metadata.namespace
```

which indeed is also the example which the [K8S documentation uses](https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/#use-pod-fields-as-values-for-environment-variables) uses.

The `KUBERNETES_NAMESPACE` is used in some of the examples in the `qiuckstarts` folder as well as in the [AsciiDoc documentation](https://github.com/eclipse/jkube/blob/master/kubernetes-maven-plugin/doc/src/main/asciidoc/inc/_introduction.adoc).

lbruun:
We assume the name prefix `KUBERNETES_` on env vars is meant for Kubernetes/OpenShift own use. Therefore we should not encourage users to create custom env vars with this name prefix.

Fixes #1132

Note:  The name `KUBERNETES_NAMESPACE` name is also used in `jkube-kit/enricher/api/src/main/java/org/eclipse/jkube/kit/enricher/handler/ContainerHandler.java` which is deliberately not touched in this PR as users may rely on this behavior ... even if it has never been documented/promised anywhere, AFAIK.

### DIFF
--- a/gradle-plugin/doc/src/main/asciidoc/inc/scenarios/_fragments.adoc
+++ b/gradle-plugin/doc/src/main/asciidoc/inc/scenarios/_fragments.adoc
@@ -26,7 +26,7 @@ The resource fragments are in `src/main/jkube`.
               - name: config
                 mountPath: /app/config
             env:
-              - name: KUBERNETES_NAMESPACE
+              - name: MY_POD_NAMESPACE
                 valueFrom:
                   fieldRef:
                     apiVersion: v1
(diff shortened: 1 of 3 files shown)


rohanKanojia (approved):

manusa (approved):
LGTM, thx!

Issue #1142, opened 2021-11-24
Log informative message whenever a goal/task is skipped


manusa:
## Description

Whenever a goal/task is skipped (in either of the plugins), we should log an informative message indicating that the execution is skipped.

- [x] Kubernetes Maven Plugin
- [x] OpenShift Maven Plugin
- [x] Kubernetes Gradle Plugin
- [x] Kubernetes Gradle Plugin

> Maybe we can create an issue to implement a log message whenever a goal (Maven) or task (Gradle) is skipped.

_Originally posted by @manusa in https://github.com/eclipse/jkube/pull/1141#discussion_r755831821_

PR #1198, opened 2022-01-07
feat(gradle-plugin): log informative message when gradle jkube tasks are skipped #1142

sunix:
## Description
part of https://github.com/eclipse/jkube/issues/1142

## Type of change

## Checklist

### DIFF
--- a/gradle-plugin/kubernetes/src/main/java/org/eclipse/jkube/gradle/plugin/task/AbstractJKubeTask.java
+++ b/gradle-plugin/kubernetes/src/main/java/org/eclipse/jkube/gradle/plugin/task/AbstractJKubeTask.java
@@ -63,8 +63,7 @@ public abstract class AbstractJKubeTask extends DefaultTask implements Kubernete
     kubernetesExtension = getProject().getExtensions().getByType(extensionClass);
   }
 
-  @TaskAction
-  public final void runTask() {
+  protected void init() {
     kubernetesExtension.javaProject = GradleUtil.convertGradleProject(getProject());
     kitLogger = createLogger(null);
     logOutputSpecFactory = new LogOutputSpecFactory(isAnsiEnabled(), kubernetesExtension.getLogStdoutOrDefault(),
@@ -88,9 +87,16 @@ public abstract class AbstractJKubeTask extends DefaultTask implements Kubernete
     } catch (IOException exception) {
       kitLogger.error("Error in fetching Build timestamps: " + exception.getMessage());
     }
-    if (canExecute()) {
-      run();
+  }
+
+  @TaskAction
+  public final void runTask() {
+    init();
+    if (!canExecute()) {
+        kitLogger.info("`%s` task is skipped.", this.getName());
+        return;
     }
+    run();
   }
 
   @Internal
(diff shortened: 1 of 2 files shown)


rohanKanojia (approved):

gradle-plugin/kubernetes/src/main/java/org/eclipse/jkube/gradle/plugin/task/AbstractJKubeTask.java
```diff
@@ -63,8 +63,7 @@ protected AbstractJKubeTask(Class<? extends KubernetesExtension> extensionClass)
     kubernetesExtension = getProject().getExtensions().getByType(extensionClass);
   }
 
-  @TaskAction
-  public final void runTask() {
+  protected void init() {
```
manusa, line 66: Any reason to extract this logic into a separate method?

rohanKanojia: I think Sun did it to make it look similar to what we have in maven plugin mojos

sunix: I did that for readability. Then we can discuss about the method names.

manusa: I guess, I didn't review those yet. But I don't understand the need for this (maybe in Maven is to be able to prepare the tests, but shouldn't be necessary).
I did give a quick glance on what was happening on Maven last week. What I noticed was that after that PR we would have execute, executeInternal, doExecute, etc. which would make it very hard to maintain.


In this case it makes it harder to read and it doesn't provide any added value such as reusability, etc. Condensing 3 lines into a method might not be that good ATM (maybe if the method grows in the future it will).

manusa: > I did that for readability. Then we can discuss about the method names.

As I said, IMHO it makes it readable when the logic contains 10s of lines or can be reused. If not, it forces you to go up and down the code  to check what's happening

sunix: Ok I moved the main method `runTask()` in the top: someone who is reading this class would see this method like a table of  content. Reader should know and understand what it is doing without having to go in the details. Then if the reader is interested to go to the details of a specific method, he can click on the method and read the content.

sunix: the second thing is to avoid having more than 1 logic in 1 method.

sunix: IMHO this is neat. Before, `runTask()` were like mainly about init stuff whereas now it is clearly 3 things:
- `init`
- `skipping or not`,
- `executing`

Selection_585

sunix: This is also raising a question: should `init` be before or after `skip`?

manusa: OK, in the GH interface it seemed that there were only 3 lines in the method.

we can keep it this way, but make init private.

manusa: > This is also raising a question: should `init` be before or after `skip`?

I guess that some `canExecute` implementations depend on the init. Even your current implementation relies on having the logger initiated.

manusa (approved):
LGTM, thx!

PR #1233, opened 2022-01-25
Fix #1232: Remove check for existing temporary tag

### LINKED ISSUE #1232: Cannot temporarily tag because target image already exists
### Description
This is the same issue that was created and fixed for docker-maven-plugin several years ago - https://github.com/fabric8io/docker-maven-plugin/issues/838
Here is the fix - https://github.com/fabric8io/docker-maven-plugin/pull/1170/files
The relevant code - https://github.com/eclipse/jkube/blob/85d9fe8fd17e59eb1cdb07da194e6419201585d1/jkube-kit/build/service/docker/src/main/java/org/eclipse/jkube/kit/build/service/docker/access/hc/DockerAccessWithHcClient.java#L661-L666

My problem is that I want to tag the image with "latest" version which already exists in the repo but pushing this new image should remove "latest" tag from an older image. This is impossible to do now.

### Info

* Eclipse JKube version : 1.5.1
* Maven version (`mvn -v`) : 3.6.2

This is my plugin configuration - using "%l" in the image name should set "latest" tag for SNAPSHOT artifacts.

            <plugin>
                <version>1.5.1</version>
                <configuration>
                    <profile>myapp</profile>
                    <access>
                        <namespace>test</namespace>
                    </access>
                    <registry>XXX.dkr.ecr.eu-west-1.amazonaws.com</registry>
                    <images>
                        <image>
                            <name>myapp:%l</name>
                            <alias>myapp</alias>
                            <build>
                                <contextDir>.</contextDir>
                            </build>
                        </image>
                    </images>
                </configuration>
            </plugin>

lagivan:
## Description

Fixes #1232 

This issue repeats the same issue from docker-maven-plugin and the proposed solution is identical - https://github.com/fabric8io/docker-maven-plugin/issues/838

The problem appears when someone wants to tag the image with "latest" version which already exists in the repo but pushing this new image should remove "latest" tag from an older image. This behavior is impossible to achieve without this fix.

### DIFF
--- a/jkube-kit/build/service/docker/src/main/java/org/eclipse/jkube/kit/build/service/docker/access/hc/DockerAccessWithHcClient.java
+++ b/jkube-kit/build/service/docker/src/main/java/org/eclipse/jkube/kit/build/service/docker/access/hc/DockerAccessWithHcClient.java
@@ -657,12 +657,6 @@ public class DockerAccessWithHcClient implements DockerAccess {
     private String tagTemporaryImage(ImageName name, String registry) throws DockerAccessException {
         String targetImage = name.getFullName(registry);
         if (!name.hasRegistry() && registry != null) {
-            if (hasImage(targetImage)) {
-                throw new DockerAccessException(
-                        String.format("Cannot temporarily tag %s with %s because target image already exists. " +
-                                        "Please remove this and retry.",
-                                name.getFullName(), targetImage));
-            }
             tag(name.getFullName(), targetImage, false);
             return targetImage;
         }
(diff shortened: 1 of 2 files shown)


manusa:
Hi @lagivan 
Your PR does remove the check, but is not a port of the solution provided for DMP https://github.com/fabric8io/docker-maven-plugin/pull/1300.

Their solution includes three modes and a specific handler for this situation.

manusa (approved):

rohanKanojia (approved):

PR #1241, opened 2022-01-26
feat: enable Create Image (pull) HTTP API options

### LINKED ISSUE #1157: Support for Docker option: --platform
### Info

* Eclipse JKube version : 1.5.1
* Maven version (`mvn -v`) : 3.8.4
* Kubernetes / Red Hat OpenShift setup and version : N/A (Not using Kubernetes, I just want to build Docker images)
* If it's a bug, how to reproduce : Not a bug

I am using a new MacBook Pro with Apple Silicon (M1 Max CPU, arm64 architecture).

I have to use some Docker base images for which there is unfortunately no native arm64 version available. This is no problem in principle; it's possible to run Docker containers for the amd64 architecture (= Intel x86_64) on this machine through Rosetta 2. (Note: Don't confuse the architecture names: arm64 = Apple Silicon, amd64 = Intel x86_64).

However: When I pull or build an image, Docker will by default try to pull or build an image for the native architecture (arm64). If no such image is available, it fails with an error message. It is possible to force Docker to use an image for the amd64 architecture by using the "--platform" option, for example: `docker pull --platform linux/amd64 tagname`

It seems that kubernetes-maven-plugin does not support supplying this option to Docker.

I tried using buildOptions as described in the documentation here: https://www.eclipse.org/jkube/docs/kubernetes-maven-plugin#jkube:build by adding this to the plugin configuration:

```
<images>
    <image>
        <name>myorg/mytag</name>
        <build>
            <buildOptions>
                <platform>linux/amd64</platform>
            </buildOptions>
        </build>
        <!-- ... -->
    </image>
</images>
```

When I run Maven with --debug, I can see that the plugin parses this configuration correctly, but it seems it does not supply it to Docker because Docker still tries to pull the base image using the wrong architecture (arm64 instead of amd64).

I also added `<apiVersion>1.41</apiVersion>` to the plugin configuration. Note that `platform` is a [supported query parameter](https://docs.docker.com/engine/api/v1.41/#operation/ImageBuild) in the Docker API v1.41.

But it still seems that kubernetes-maven-plugin does not pass this parameter to Docker.

Please make the plugin support this option, so that I can use it to build amd64 Docker images on my Apple Silicon MacBook Pro.

manusa:
## Description

Fix #1157 

## Type of change

## Checklist
 - [x] I tested my code in Kubernetes

### DIFF
--- a/kubernetes-maven-plugin/doc/src/main/asciidoc/inc/build/_configuration.adoc
+++ b/kubernetes-maven-plugin/doc/src/main/asciidoc/inc/build/_configuration.adoc
@@ -22,6 +22,13 @@ described in <<build-buildargs,Build Args>>
 https://docs.docker.com/engine/reference/api/docker_remote_api_v1.24/#build-image-from-a-dockerfile[Docker Remote API] and are restricted to simple options
 (e.g.: memory, shmsize). If you use the respective configuration options for build options natively supported by the build configuration (i.e. `noCache`, `cleanup=remove` for buildoption `forcerm=1` and `args` for build args) then these will override any corresponding options given here. The key-value syntax is the same as when defining environment variables or labels as described in <<misc-env,Setting Environment Variables and Labels>>.
 
+| *createImageOptions*
+| Map specifying the create image options to provide to the docker daemon when pulling or
+  importing an image.
+  These options map to the ones listed as query parameters in the
+  https://docs.docker.com/engine/api/v1.41/#operation/ImageCreate[Docker Remote API] and are
+  restricted to simple options (e.g.: fromImage, fromSrc, platform).
+
 | *cleanup*
 | Cleanup dangling (untagged) images after each build (including any containers created from them). Default is `try` which tries to remove the old image, but doesn't fail the build if this is not possible because e.g. the image is still used by a running container. Use `remove` if you want to fail the build and `none` if no cleanup is requested.
 
--- a/quickstarts/maven/xml-config/pom.xml
+++ b/quickstarts/maven/xml-config/pom.xml
@@ -174,7 +174,7 @@
               <!-- "alias" is used to correlate to the containers in the pod spec -->
               <alias>camel-service</alias>
               <build>
-                <from>quay.io/jkube/jkube-java:0.0.13</from>
+                <from>quay.io/jkube/jkube-java-11:0.0.13</from>
                 <assembly>
                   <targetDir>/deployments</targetDir>
                   <inline>
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ target/
 build/
 !**/src/main/**/build/
 !**/src/test/**/build/
+!jkube-kit/build/
 *.iml
 *.ipr
 *.iws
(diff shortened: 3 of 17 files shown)


kubernetes-maven-plugin/doc/src/main/asciidoc/inc/build/_configuration.adoc
```diff
@@ -22,6 +22,13 @@ described in <<build-buildargs,Build Args>>
 https://docs.docker.com/engine/reference/api/docker_remote_api_v1.24/#build-image-from-a-dockerfile[Docker Remote API] and are restricted to simple options
 (e.g.: memory, shmsize). If you use the respective configuration options for build options natively supported by the build configuration (i.e. `noCache`, `cleanup=remove` for buildoption `forcerm=1` and `args` for build args) then these will override any corresponding options given here. The key-value syntax is the same as when defining environment variables or labels as described in <<misc-env,Setting Environment Variables and Labels>>.
 
+| *createImageOptions*
```
rohanKanojia, line 25: This might be just my opinion. I think createImageOptions might be confusing from an end-user perspective. User also creates images of this project using build process. In my opinion it should be named `pullOptions`/ `pullImageOptions`, `importOptions` etc

rohanKanojia (approved):

kubernetes-maven-plugin/doc/src/main/asciidoc/inc/build/_configuration.adoc
manusa: This is named according to the HTTP API exposed by Docker.

I agree it might be confusing, but just as `buildOptions`. Both these configurations are solely in the scope of the docker daemon and its HTTP API. IMHO making them as specific as possible, makes it harder for users to confuse them with something else.

However, I don't feel there is a better way to clear confusion. `pullOptions` or `importOptions` are confusing too i.e. what is the purpose of these options? Is this something generic? are these options applied when using S2I/JIB?

Considering a full example:
```xml
<build>
  <createImageOptions>
    <platform>linux</platform>
    <message>This is a docker commit message</message>
  </createImageOptions>
</build>
```

IMO, since this is something very specific to Docker, it's better to keep the name as specific as possible in order to prevent users having other expectations.

rohanKanojia: ohk, let's keep it.

sunix (approved):
I added a minor question but looks good to me

quickstarts/maven/xml-config/pom.xml
```diff
@@ -174,7 +174,7 @@
               <!-- "alias" is used to correlate to the containers in the pod spec -->
               <alias>camel-service</alias>
               <build>
-                <from>quay.io/jkube/jkube-java:0.0.13</from>
+                <from>quay.io/jkube/jkube-java-11:0.0.13</from>
```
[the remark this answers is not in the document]
manusa: Not really, this is a fix for this example (won't work with Java17).
I can commit the change in a separate commit. It might be better.

PR #1281, opened 2022-02-14
fix (jkube-kit) : Add default restartPolicy in case of controller type Job (#1278)

### LINKED ISSUE #1278: Enricher with type Job does not generate mandatory spec.template.spec.restartPolicy
### Description
If you configure 

`<jkube.enricher.jkube-controller.type>job</jkube.enricher.jkube-controller.type>
`

, do not set any additional YAML fragment for the Job, and run `mvn k8s:resource`, the resulted Job resource will not have a `spec.template.spec.restartPolicy`. When trying to apply the resource (`mvn k8s:apply`) you will get a message with 

`This is required Policy: Required value: valid values: "OnFailure", "Never". `

### Info

* Eclipse JKube version : 1.6.0
* Maven version (`mvn -v`) :3.8.4
```

```
* Kubernetes / Red Hat OpenShift setup and version : Kubernetes v1.23/ Red Hat OpenShift 4.8

* If it's a bug, how to reproduce : Do not use any **Job fragment** under **src/main/jkube** and run mvn k8s:resource.

* Sample Reproducer Project : *[GitHub Clone URL]*https://github.com/ammbra/cat-adoption-batch. To reproduce it, please use the jkube branch https://github.com/ammbra/cat-adoption-batch/tree/jkube

rohanKanojia:
## Description
Fix #1278

Right now JKube doesn't generate resources with pod spec's
`restartPolicy` set. This works for most of the controller type
resources (e.g `Deployment`, `StatefulSet`, `DaemonSet`, `ReplicaSet`,
`ReplicationController` ); If not provided Kubernetes assumes it to be
`Always`.

But in case of `Job` it doesn't work and Kubernetes gives `422`. Add a
default restart policy in case of `Job`.

[kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#restart-policy](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#restart-policy)

## Type of change

## Checklist

### DIFF
--- a/jkube-kit/enricher/generic/src/main/java/org/eclipse/jkube/enricher/generic/DefaultControllerEnricher.java
+++ b/jkube-kit/enricher/generic/src/main/java/org/eclipse/jkube/enricher/generic/DefaultControllerEnricher.java
@@ -119,12 +119,13 @@ public class DefaultControllerEnricher extends BaseEnricher {
   public void create(PlatformMode platformMode, KubernetesListBuilder builder) {
     final String name = getConfig(Config.NAME,
         JKubeProjectUtil.createDefaultResourceName(getContext().getGav().getSanitizedArtifactId()));
-    ResourceConfig xmlResourceConfig = Optional.ofNullable(getConfiguration().getResource())
+    ResourceConfig providedResourceConfig = Optional.ofNullable(getConfiguration().getResource())
         .orElse(ResourceConfig.builder().build());
-    ResourceConfig config = ResourceConfig.toBuilder(xmlResourceConfig)
-        .controllerName(getControllerName(xmlResourceConfig, name))
-        .imagePullPolicy(getImagePullPolicy(xmlResourceConfig, getConfig(Config.PULL_POLICY)))
-        .replicas(getReplicaCount(builder, xmlResourceConfig, Configs.asInt(getConfig(Config.REPLICA_COUNT))))
+    ResourceConfig config = ResourceConfig.toBuilder(providedResourceConfig)
+        .controllerName(getControllerName(providedResourceConfig, name))
+        .imagePullPolicy(getImagePullPolicy(providedResourceConfig, getConfig(Config.PULL_POLICY)))
+        .replicas(getReplicaCount(builder, providedResourceConfig, Configs.asInt(getConfig(Config.REPLICA_COUNT))))
+        .restartPolicy(providedResourceConfig.getRestartPolicy())
         .build();
 
     final List<ImageConfiguration> images = getImages();
@@ -144,5 +145,4 @@ public class DefaultControllerEnricher extends BaseEnricher {
   private static List<String> getContainersFromPodSpec(PodTemplateSpec spec) {
     return spec.getSpec().getContainers().stream().map(Container::getName).collect(Collectors.toList());
   }
-
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ Usage:
 * Fix #1236: Add integration tests for DefaultControllerEnricher in gradle plugins
 * Fix #1237: Add gradle project integration test for DefaultMetadataEnricher
 * Fix #1240: Add documentation for DependencyEnricher
+* Fix #1278: Enricher with type Job does not generate mandatory `spec.template.spec.restartPolicy`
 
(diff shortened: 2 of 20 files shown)


jkube-kit/enricher/api/src/main/java/org/eclipse/jkube/kit/enricher/api/BaseEnricher.java
```diff
@@ -55,6 +55,7 @@
     public static final String CREATE_EXTERNAL_URLS = "jkube.createExternalUrls";
     public static final String JKUBE_DOMAIN = "jkube.domain";
     public static final String JKUBE_ENFORCED_REPLICAS = "jkube.replicas";
+    public static final String DEFAULT_JOB_RESTART_POLICY = "OnFailure";
```
manusa, line 58: Was this the default behavior when it was an optional field? do you have any reference?

rohanKanojia: As pointed out in PR description, the default value is `Always` (if not specified) for most controllers. However, for `Job` it's mandatory to provide restart policy

https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#restart-policy

jkube-kit/enricher/generic/src/main/java/org/eclipse/jkube/enricher/generic/DefaultControllerEnricher.java
```diff
@@ -145,4 +146,10 @@ public void create(PlatformMode platformMode, KubernetesListBuilder builder) {
     return spec.getSpec().getContainers().stream().map(Container::getName).collect(Collectors.toList());
   }
 
+  private static String getDefaultContainerRestartPolicy(String controllerType) {
+    if (controllerType != null && controllerType.equalsIgnoreCase(ControllerType.JOB.name())) {
+      return DEFAULT_JOB_RESTART_POLICY;
+    }
+    return null;
+  }
```
manusa, line 154: this should be delegated to the handler (it makes no sense having an if-else condition here, given that we have specific handlers to take care of specifics).

rohanKanojia: JobHandler seems to be generating pod template spec using PodTemplateHandler:
https://github.com/eclipse/jkube/blob/b4c31fe43d6ee853bcf85128eb31727176a9d258/jkube-kit/enricher/api/src/main/java/org/eclipse/jkube/kit/enricher/handler/JobHandler.java#L67-L70

We don't have any way of knowing whether PodTemplate is generated for Job or any other kind of controller in PodTemplateHandler. That's why I'm setting restart policy from DefaultControllerEnricher from where handlers are invoked.

jkube-kit/enricher/api/src/main/java/org/eclipse/jkube/kit/enricher/api/BaseEnricher.java
manusa: But what was the behavior for Job controllers when this value wasn't required. That's the question

manusa: Right now, we're assuming it is `OnFailure`.

rohanKanojia: It wasn't provided and Job was created without any `restartPolicy`. This used to get rejected by Kubernetes api server on apply phase

manusa: I understand the issue. What I mean, is that I assumed that at some point this was working (however I'm doubting this more and more), so the kube API changed the requirements for Job Pod Templates at some point.

manusa: OK, so my worries are confirmed, this feature never worked...

https://github.com/kubernetes/kubernetes/issues/41194

manusa (approved):

Issue #1324, opened 2022-03-03
should pick tomcat 10 if Jakarta EE provided api is used in the dependencies


sunix:
### Description
For webapp using jakarta ee, tomcat 10 should be used instead of 9

```
  <dependencies>
    <!-- Add servlet dependencies -->
    <dependency>
      <groupId>jakarta.servlet</groupId>
      <artifactId>jakarta.servlet-api</artifactId>
      <version>5.0.0</version>
      <scope>provided</scope>
    </dependency>
    <!-- Add jsp dependencies -->
    <dependency>
      <groupId>jakarta.servlet.jsp</groupId>
      <artifactId>jakarta.servlet.jsp-api</artifactId>
      <version>3.0.0</version>
      <scope>provided</scope>
    </dependency>
``` 

```
import jakarta.servlet.ServletException;
import jakarta.servlet.http.HttpServlet;
import jakarta.servlet.http.HttpServletRequest;
import jakarta.servlet.http.HttpServletResponse;
```

### Info

* Eclipse JKube version : 1.7.0
* If it's a bug, how to reproduce : I faced that while working on the demo https://github.com/sunix/hello-tomcat-kubernetes
* Sample Reproducer Project : *[GitHub Clone URL]*

manusa:
We need to create a base Tomcat10 image.

Tomcat 10 should be used in all cases or if we want to keep Tomcat9 too, implement a new Handler for the new conditions.

PR #1385, opened 2022-03-23
fix(tomcat-webapp-generator): Support legacy javaee as well as jakartaee projects

sunix:
## Description
~~This is picking the right base image for tomcat if legacy javaee dependencies are used in the project~~
- by default, war file is copied to the compat/migrate directory webapps-javaee (regardless of its dependencies)
- if deployment directory is webapps-javaee, INFO message is printed indicating this and how to make the deployment faster if already jakartaee compliant, by setting the deployment folder to webapps.
- Users should be able to provide a custom configuration and rely on our Tomcat 9 image as a last resort workaround.

Fix https://github.com/eclipse/jkube/issues/1324
Depends on https://github.com/jkubeio/jkube-images/pull/15

## Type of change

## Checklist

### DIFF
--- a/jkube-kit/doc/src/main/asciidoc/inc/generator/_webapp.adoc
+++ b/jkube-kit/doc/src/main/asciidoc/inc/generator/_webapp.adoc
@@ -33,8 +33,8 @@ The base images chosen are:
 | | Docker Build | S2I Build
 
 | *Tomcat*
-| `quay.io/jkube/jkube-tomcat9`
-| `quay.io/jkube/jkube-tomcat9`
+| `quay.io/jkube/jkube-tomcat`
+| `quay.io/jkube/jkube-tomcat`
 
 | *Jetty*
 | `quay.io/jkube/jkube-jetty9`
--- a/jkube-kit/doc/src/main/asciidoc/inc/generator/_webapp.adoc
+++ b/jkube-kit/doc/src/main/asciidoc/inc/generator/_webapp.adoc
@@ -9,7 +9,7 @@ endif::[]
 ifeval::["{plugin-type}" == "maven"]
 The `webapp` generator tries to detect WAR builds and selects a base servlet container image based on the configuration found in the `pom.xml`:
 
-* A **Tomcat** base image is selected when a `tomcat6-maven-plugin`,
+* A **Tomcat** base image is selected by default,
 `tomcat7-maven-plugin` or `tomcat8-maven-plugin` is present or when a `META-INF/context.xml` could be found in the classes directory.
 * A **Jetty** base image is selected when a `jetty-maven-plugin` is present or one of the files `WEB-INF/jetty-web.xml` or `WEB-INF/jetty-logging.properties` is found.
 * A **Wildfly** base image is chosen for a given `jboss-as-maven-plugin` or `wildfly-maven-plugin` or when a Wildfly specific deployment descriptor like `jboss-web.xml` is found.
@@ -91,3 +91,18 @@ This may be required for Wildfly webapp s2i build to compose a `WildFly` server 
 endif::[]
 | `jkube.generator.webapp.env`
 |===
+
+
+==== JakartaEE and retrocompatibility with JavaEE in Tomcat
+From Tomcat 10, only JakartaEE compliant projects are supported. However, legacy JavaEE projects can automatically be migrated by deploying the war in `${CATALINA_HOME}/webapps-javaee`. By default, the webapp generator is based on a Tomcat 10+ image and will copy the war file to `${CATALINA_HOME}/webapps-javaee`.
+
+If the project is already JakartaEE compliant, it is recommanded to set the webapp directory to `${CATALINA_HOME}/webapps`. This can be done by setting the property `jkube.generator.webapp.env` to `TOMCAT_WEBAPPS_DIR=webapps`.
+
+To keep using Tomcat 9, set the properties:
+
+* `jkube.generator.webapp.from` to `quay.io/jkube/jkube-tomcat9:0.0.15`
+* `jkube.generator.webapp.cmd` to `/usr/local/s2i/run`
+ifdef::doc-openshift[]
+* `jkube.generator.webapp.supportsS2iBuild` to `true`
+endif::[]
+
--- a/jkube-kit/generator/webapp/src/main/java/org/eclipse/jkube/generator/webapp/handler/TomcatAppSeverHandler.java
+++ b/jkube-kit/generator/webapp/src/main/java/org/eclipse/jkube/generator/webapp/handler/TomcatAppSeverHandler.java
@@ -14,6 +14,8 @@
 package org.eclipse.jkube.generator.webapp.handler;
 
 import java.io.IOException;
+import java.util.Collections;
+import java.util.Map;
 import java.util.stream.Stream;
 
 import org.eclipse.jkube.generator.api.GeneratorContext;
@@ -64,5 +66,15 @@ public class TomcatAppSeverHandler extends AbstractAppServerHandler {
   @Override
   public boolean supportsS2iBuild() {
     return true;
+  }
+
+  @Override
+  public Map<String, String> getEnv() {
+    // Setting Tomcat webapps dir to `webapps-javaee` by default for
+    // retrocompatibility.
+    // If the project is already JakartaEE compliant, user should override
+    // `jkube.generator.webapp.env` to `TOMCAT_WEBAPPS_DIR=webapps` for a faster
+    // startup.
+    return Collections.singletonMap("TOMCAT_WEBAPPS_DIR", "webapps-javaee");
   }
 }
--- a/jkube-kit/parent/pom.xml
+++ b/jkube-kit/parent/pom.xml
@@ -70,7 +70,7 @@
     <version.wagon-ssh-external>2.3</version.wagon-ssh-external>
 
     <!-- =======================================================  -->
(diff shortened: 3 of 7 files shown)


sunix:
@rohanKanojia nice :) the preview is working https://13973-216815773-gh.circle-artifacts.com/0/docs-generated/kubernetes-maven-plugin.html#generator-webapp

jkube-kit/common/src/test/java/org/eclipse/jkube/kit/common/util/JKubeProjectUtilTest.java
```diff
@@ -46,41 +46,94 @@ public void hasDependencyWithGroupIdWithNulls() {
...
+            Dependency.builder().groupId("javax.servlet").artifactId("javax.servlet-api").version("3.1.0").build(), //
+            Dependency.builder().groupId("io.other").artifactId("artifact").version("1.3.37").build() //
+        ));
+    // When
+    final boolean result = JKubeProjectUtil.hasDependencyWithGroupIdMatching(project, JAVAEE_GROUPID_REGEXP);
+    // Then
+    assertThat(result).isTrue();
+  }
+
+  static final String JAVAEE_GROUPID_REGEXP = "javax([/\\.](activation" + //
```
rohanKanojia, line 96: nit, could this be moved to top. Cpuld this be private?

rohanKanojia (approved):

jkube-kit/common/src/test/java/org/eclipse/jkube/kit/common/util/JKubeProjectUtilTest.java
sunix: I made this private
But I feel like it should be kept here just after the first reference to it: to avoid readers to scroll up and down.

jkube-kit/generator/webapp/src/main/java/org/eclipse/jkube/generator/webapp/handler/TomcatAppSeverHandler.java
```diff
@@ -48,9 +48,41 @@ private boolean hasTomcatMavenPlugin() {
 
   @Override
   public String getFrom() {
+    if (projectHasLegacyJavaeeDependencies()) {
+      return imageLookup.getImageName("tomcat9.upstream.docker");
+    }
     return imageLookup.getImageName("tomcat.upstream.docker");
   }
 
+  private boolean projectHasLegacyJavaeeDependencies() {
```
manusa, line 57: I'm not really following the logic behind this (especially considering the kind of dependencies you list). Plus, this is not aligned with the issue description (it's the complete opposite).

Maybe we should start by clarifying why Tomcat 10 might not be applicable for "legacy" apps and in which cases we should fallback to 9 (or vice versa, as the issue suggests). In the issue, the opposite situation was described, a dependency towards a "newer" artifact made the project incompatible with Tomcat 9.

AFAIK Tomcat 10 is breaking because it uses jakarta implementations of the javax packaged APIs (javax ---> jakarta). However, a dependency towards the javax.mail API jar, doesn't imply that your project won't have a dependency to the implementation jar too (e.g. `com.sun.mail:javax.mail`), thus making it viable for Tomcat 10.

sunix: Sorry, I did not mentioned where this was coming from, maybe it should be specified in a comment.
https://github.com/apache/tomcat-jakartaee-migration#differences-between-java-ee-8-and-jakarta-ee-9

You are absolutely right, we would need to check the usage of the import packages in the generated war.

For `com.sun.mail:javax.mail` it would work by chance because it has `javax.activation:activation` as a transitive dependency

manusa: > For `com.sun.mail:javax.mail` it would work by chance because it has `javax.activation:activation` as a transitive dependency

What I'm trying to tell is that you can deploy into Tomcat10 webapp server any javax API based application if you include the implementation libraries too.

So maybe, instead of making this something `smart` we might just want to make it configurable.

sunix: OK what about we just log a warning with the same condition here and suggest the user to use `jkube.generator.webapp.server=tomcat9` if any issue ?

manusa: The main problem I see is that your condition is checking on group IDs and not on package imports.
Maybe we can just print an INFO message stating that Tomcat 10 was selected, and suggest the configuration property with a message like "If your project is still using javax.* imports..."

sunix: This is interesting: 

> Tomcat can convert an existing web application from Java EE 8 to Jakarta EE 9 at deployment time using the Apache Tomcat migration tool for Jakarta EE. To make use of the feature, the web application should be placed in the Host legacyAppBase folder (by default named webapps-javaee) and they will be converted to an equivalent Jakarta EE web application in the Host appBase folder (by default named webapps).

from https://tomcat.apache.org/migration-10.html#Migrating_from_9.0.x_to_10.0.x

sunix: > The main problem I see is that your condition is checking on group IDs and not on package imports. Maybe we can just print an INFO message stating that Tomcat 10 was selected, and suggest the configuration property with a message like "If your project is still using javax.* imports..."

I know it is not perfect. But this would give a good hint in 99% of the case.  Whereas if we show the message all the time, it would be misleading.
As a user I would prefer, everything automated with the log messages explaining the choices that had been made by JKube and the alternative possible options. So basically keep this condition that would detect the 99% right configuration with a log message that the webapp had been placed to `webapps-javaee` and that it is possible to place it in `webapp` if it doesn't fit.

manusa: To summarize:
- war file is copied to the compat/migrate directory `webapps-javaee` (regardless of its dependencies) ---> Test if this works
- INFO message is printed indicating this, plus suggesting the config flag in case the application fails to start

manusa: So if I understand correctly, we provide 2 configurations, one to set the tomcat version and another to set the deployment directory.
If any of this configurations is the default, then we log the INFO message on why this was chosen and how to override the configuration.
Right?

sunix: I think we can drop the tomcat version. setting the deployment directory is good enough. Do we still need to keep `tomcat9` beside the previous JKube versions ?

sunix:
I have just pushed new changes to reflect our previous discussion.
It depends on https://github.com/jkubeio/jkube-images/pull/15
If the approach is ok, i'll update the doc regarding this PR.

jkube-kit/doc/src/main/asciidoc/inc/generator/_webapp.adoc
```diff
@@ -91,3 +91,18 @@ This may be required for Wildfly webapp s2i build to compose a `WildFly` server
...
+
+
+==== JakartaEE and retrocompatibility with JavaEE in Tomcat
+From Tomcat 10, only JakartaEE compliant projects are supported. However, legacy JavaEE projects can automatically be migrated by deploying the war in `${CATALINA_HOME}/webapps-javaee`. By default, the webapp generator is based on a Tomcat 10+ image and will copy the war file to `${CATALINA_HOME}/webapps-javaee`.
+
+If the project is already JakartaEE compliant, it is recommanded to set the webapp directory to `${CATALINA_HOME}/webapps`. This can be done by setting the property `jkube.generator.webapp.env` to `TOMCAT_WEBAPPS_DIR=webapps`.
+
+To keep using Tomcat 9, set the properties:
+
+* `jkube.generator.webapp.from` to `quay.io/jkube/jkube-tomcat9:0.0.14`
```
manusa, line 103:
```suggestion
* `jkube.generator.webapp.from` to `quay.io/jkube/jkube-tomcat9:0.0.15`
```

jkube-kit/parent/pom.xml
```diff
@@ -109,7 +109,7 @@
     <image.karaf.upstream.istag>jkube-karaf:${version.image.karaf.upstream.s2i}</image.karaf.upstream.istag>
 
     <!-- webapp -->
-    <image.tomcat.upstream>quay.io/jkube/jkube-tomcat9:${version.image.tomcat.upstream}</image.tomcat.upstream>
+    <image.tomcat.upstream>quay.io/jkube/jkube-tomcat:${version.image.tomcat.upstream}</image.tomcat.upstream>
```
manusa, line 112: `version.image.jkube-images` value needs to be upgraded to 0.0.15 too

manusa (changes requested):
My last two comments should be addressed to make everything work.

It would also be nice to include (in a follow-up) PR some sort of test to verify the configuration options, etc.

manusa (approved):

Issue #1474, opened 2022-04-25
WildflyJARHealthCheckEnricher : Add `/health/started` as startup probe if enabled


rohanKanojia:
### Description

Blocked by https://github.com/eclipse/jkube/issues/1322

[Wildfly Microprofile Health Subsystem](https://github.com/wildfly/wildfly/blob/main/docs/src/main/asciidoc/_admin-guide/subsystem-configuration/MicroProfile_Health.adoc#http-endpoints) configuration defines three HTTP endpoints:

- `/health/live` to test the liveness of the application server

- `/health/ready` to test the readiness of the application server.

- `/health/started` to test the startup of the application server.

WildflyJARHealthCheckEnricher seems to be adding `/health/live` and `/health/ready` for liveness and readiness probes. We need to add support for startup probe in that as well.

- [x] Add `/health/started` as startup probe in WildflyJARHealthCheckEnricher
- [x] UpdateWildflyJARHealthCheckEnricher  documentation about new probe

anurag-rajawat:
@rohanKanojia may I work on this one?

PR #1492, opened 2022-05-09
feat (jkube-kit/wildfly-jar): Add startup probe in WildFlyJARHealthCheckEnricher (#1474)

anurag-rajawat:
## Description
Fix: #1474

Signed-off-by: Anurag Singh Rajawat <anuragsinghrajawat22@gmail.com>

## Type of change

## Checklist

### DIFF
--- a/jkube-kit/jkube-kit-wildfly-jar/src/main/java/org/eclipse/jkube/wildfly/jar/enricher/WildflyJARHealthCheckEnricher.java
+++ b/jkube-kit/jkube-kit-wildfly-jar/src/main/java/org/eclipse/jkube/wildfly/jar/enricher/WildflyJARHealthCheckEnricher.java
@@ -19,28 +19,34 @@ import io.fabric8.kubernetes.api.model.EnvVarBuilder;
 import io.fabric8.kubernetes.api.model.KubernetesListBuilder;
 import io.fabric8.kubernetes.api.model.Probe;
 import io.fabric8.kubernetes.api.model.ProbeBuilder;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Map;
 import lombok.AllArgsConstructor;
 import lombok.Getter;
 import org.eclipse.jkube.kit.common.Configs;
-import org.eclipse.jkube.kit.enricher.api.JKubeEnricherContext;
-import org.eclipse.jkube.kit.enricher.specific.AbstractHealthCheckEnricher;
-
 import org.eclipse.jkube.kit.common.JavaProject;
 import org.eclipse.jkube.kit.common.Plugin;
 import org.eclipse.jkube.kit.common.util.JKubeProjectUtil;
 import org.eclipse.jkube.kit.config.resource.PlatformMode;
+import org.eclipse.jkube.kit.enricher.api.JKubeEnricherContext;
+import org.eclipse.jkube.kit.enricher.specific.AbstractHealthCheckEnricher;
+
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+
+import static org.eclipse.jkube.kit.common.Configs.asBoolean;
+import static org.eclipse.jkube.kit.common.Configs.asInt;
+import static org.eclipse.jkube.wildfly.jar.WildflyJarUtils.BOOTABLE_JAR_ARTIFACT_ID;
+import static org.eclipse.jkube.wildfly.jar.WildflyJarUtils.BOOTABLE_JAR_GROUP_ID;
+import static org.eclipse.jkube.wildfly.jar.WildflyJarUtils.DEFAULT_LIVENESS_PATH;
+import static org.eclipse.jkube.wildfly.jar.WildflyJarUtils.DEFAULT_READINESS_PATH;
+import static org.eclipse.jkube.wildfly.jar.WildflyJarUtils.DEFAULT_STARTUP_PATH;
+import static org.eclipse.jkube.wildfly.jar.WildflyJarUtils.isStartupEndpointSupported;
 
 /**
  * Enriches wildfly-jar containers with health checks if the bootable JAR has
  * been configured for cloud.
  */
 public class WildflyJARHealthCheckEnricher extends AbstractHealthCheckEnricher {
-
-    public static final String BOOTABLE_JAR_GROUP_ID = "org.wildfly.plugins";
-    public static final String BOOTABLE_JAR_ARTIFACT_ID = "wildfly-jar-maven-plugin";
 
     public WildflyJARHealthCheckEnricher(JKubeEnricherContext buildContext) {
         super(buildContext, "jkube-healthcheck-wildfly-jar");
@@ -51,13 +57,16 @@ public class WildflyJARHealthCheckEnricher extends AbstractHealthCheckEnricher {
 
         SCHEME("scheme", "HTTP"),
         PORT("port", "9990"),
-        FAILURETHRESHOLD("failureThreshold","3"),
-        SUCCESSTHRESHOLD("successThreshold","1"),
-        LIVENESSINITIALDELAY("livenessInitialDelay", "60"),
-        READINESSINITIALDELAY("readinessInitialDelay", "10"),
-        READINESSPATH("readinessPath", "/health/ready"),
-        LIVENESSPATH("livenessPath", "/health/live"),
-        ENFORCEPROBES("enforceProbes","false");
+        FAILURE_THRESHOLD("failureThreshold","3"),
+        SUCCESS_THRESHOLD("successThreshold","1"),
+        LIVENESS_INITIAL_DELAY("livenessInitialDelay", "60"),
+        READINESS_INITIAL_DELAY("readinessInitialDelay", "10"),
+        STARTUP_INITIAL_DELAY("startupInitialDelay", "10"),
+        READINESS_PATH("readinessPath", DEFAULT_READINESS_PATH),
+        LIVENESS_PATH("livenessPath", DEFAULT_LIVENESS_PATH),
+        STARTUP_PATH("startupPath", DEFAULT_STARTUP_PATH),
+        PERIOD_SECONDS("periodSeconds", "10"),
+        ENFORCE_PROBES("enforceProbes","false");
 
         @Getter
         protected String key;
@@ -67,12 +76,40 @@ public class WildflyJARHealthCheckEnricher extends AbstractHealthCheckEnricher {
 
     @Override
     protected Probe getReadinessProbe() {
-        return discoverWildflyJARHealthCheck(true);
+        return discoverWildflyJARHealthCheck(Config.READINESS_PATH, Config.READINESS_INITIAL_DELAY);
     }
 
     @Override
     protected Probe getLivenessProbe() {
-        return discoverWildflyJARHealthCheck(false);
+        return discoverWildflyJARHealthCheck(Config.LIVENESS_PATH, Config.LIVENESS_INITIAL_DELAY);
+    }
+
+    @Override
+    protected Probe getStartupProbe() {
+      if (isStartupEndpointSupported(getContext().getProject())) {
+          return discoverWildflyJARHealthCheck(Config.STARTUP_PATH, Config.STARTUP_INITIAL_DELAY);
+      }
+      return null;
+    }
+
+    private Probe discoverWildflyJARHealthCheck(Config path, Config initialDelay) {
+        if (isAvailable()) {
+            int port = asInt(getConfig(Config.PORT));
+            if (port <= 0) {
+                return null;
+            }
+            return new ProbeBuilder()
+                    .withNewHttpGet().withNewPort(port)
+                    .withPath(getConfig(path))
+                    .withScheme(getConfig(Config.SCHEME).toUpperCase())
+                    .endHttpGet()
+                    .withFailureThreshold(asInt(getConfig(Config.FAILURE_THRESHOLD)))
+                    .withSuccessThreshold(asInt(getConfig(Config.SUCCESS_THRESHOLD)))
+                    .withInitialDelaySeconds(asInt(getConfig(initialDelay)))
+                    .withPeriodSeconds(asInt(getConfig(Config.PERIOD_SECONDS)))
+                    .build();
+        }
+        return null;
     }
 
     @Override
@@ -100,29 +137,8 @@ public class WildflyJARHealthCheckEnricher extends AbstractHealthCheckEnricher {
         super.create(platformMode, builder);
     }
 
-    private Probe discoverWildflyJARHealthCheck(boolean isReadiness) {
-
-        if (isAvailable()) {
-            Integer port = getPort();
-            if (port <= 0) {
-                return null;
-            }
-            int initialDelay = isReadiness ? getReadinessInitialDelay() : getLivenessInitialDelay();
-            // scheme must be in upper case in k8s
-            String scheme = getScheme().toUpperCase();
-            String path = isReadiness ? getReadinessPath() : getLivenessPath();
-
-            return new ProbeBuilder()
-                    .withNewHttpGet().withNewPort(port).withPath(path).withScheme(scheme).endHttpGet()
-                    .withFailureThreshold(getFailureThreshold())
-                    .withSuccessThreshold(getSuccessThreshold())
-                    .withInitialDelaySeconds(initialDelay).build();
-        }
-        return null;
-    }
-
     private boolean isAvailable() {
-        if (isProbeEnforced()) {
+        if (asBoolean(getConfig(Config.ENFORCE_PROBES))) {
             return true;
         }
         JavaProject project = getContext().getProject();
@@ -134,39 +150,4 @@ public class WildflyJARHealthCheckEnricher extends AbstractHealthCheckEnricher {
         return config.containsKey("cloud");
     }
 
-    protected int getFailureThreshold() {
-        return Configs.asInteger(getConfig(Config.FAILURETHRESHOLD));
-    }
-
-    protected boolean isProbeEnforced() {
-        return Configs.asBoolean(getConfig(Config.ENFORCEPROBES));
-    }
-
-    protected int getSuccessThreshold() {
-        return Configs.asInteger(getConfig(Config.SUCCESSTHRESHOLD));
-    }
-
-    protected String getScheme() {
-        return Configs.asString(getConfig(Config.SCHEME));
-    }
-
-    protected int getPort() {
-        return Configs.asInt(getConfig(Config.PORT));
-    }
-
-    protected String getLivenessPath() {
-        return Configs.asString(getConfig(Config.LIVENESSPATH));
-    }
-
-    protected int getLivenessInitialDelay() {
-        return Configs.asInt(getConfig(Config.LIVENESSINITIALDELAY));
-    }
-
-    protected String getReadinessPath() {
-        return Configs.asString(getConfig(Config.READINESSPATH));
-    }
-
-    protected int getReadinessInitialDelay() {
-        return Configs.asInt(getConfig(Config.READINESSINITIALDELAY));
-    }
 }
--- a/jkube-kit/jkube-kit-wildfly-jar/src/main/java/org/eclipse/jkube/wildfly/jar/generator/WildflyJARGenerator.java
+++ b/jkube-kit/jkube-kit-wildfly-jar/src/main/java/org/eclipse/jkube/wildfly/jar/generator/WildflyJARGenerator.java
@@ -13,26 +13,27 @@
  */
 package org.eclipse.jkube.wildfly.jar.generator;
 
+import org.eclipse.jkube.generator.api.GeneratorContext;
+import org.eclipse.jkube.generator.javaexec.JavaExecGenerator;
+import org.eclipse.jkube.kit.common.AssemblyFileSet;
+import org.eclipse.jkube.kit.common.JavaProject;
+import org.eclipse.jkube.kit.common.Plugin;
+import org.eclipse.jkube.kit.common.util.JKubeProjectUtil;
+import org.eclipse.jkube.kit.config.image.ImageConfiguration;
+
 import java.io.File;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
-import org.eclipse.jkube.generator.api.GeneratorContext;
-import org.eclipse.jkube.generator.javaexec.JavaExecGenerator;
-import org.eclipse.jkube.kit.config.image.ImageConfiguration;
-import org.eclipse.jkube.kit.common.util.JKubeProjectUtil;
-
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
-import org.eclipse.jkube.kit.common.AssemblyFileSet;
-import org.eclipse.jkube.kit.common.JavaProject;
-import org.eclipse.jkube.kit.common.Plugin;
-import org.eclipse.jkube.wildfly.jar.enricher.WildflyJARHealthCheckEnricher;
-import static org.eclipse.jkube.wildfly.jar.enricher.WildflyJARHealthCheckEnricher.BOOTABLE_JAR_ARTIFACT_ID;
-import static org.eclipse.jkube.wildfly.jar.enricher.WildflyJARHealthCheckEnricher.BOOTABLE_JAR_GROUP_ID;
 
+import static org.eclipse.jkube.wildfly.jar.WildflyJarUtils.BOOTABLE_JAR_ARTIFACT_ID;
+import static org.eclipse.jkube.wildfly.jar.WildflyJarUtils.BOOTABLE_JAR_GROUP_ID;
+
+@SuppressWarnings("unchecked")
 public class WildflyJARGenerator extends JavaExecGenerator {
     static final String JBOSS_MAVEN_DIST = "jboss-maven-dist";
     static final String JBOSS_MAVEN_REPO = "jboss-maven-repo";
@@ -56,8 +57,7 @@ public class WildflyJARGenerator extends JavaExecGenerator {
     @Override
     public boolean isApplicable(List<ImageConfiguration> configs) {
         return shouldAddGeneratedImageConfiguration(configs)
-                && JKubeProjectUtil.hasPlugin(getProject(),
-                        WildflyJARHealthCheckEnricher.BOOTABLE_JAR_GROUP_ID, WildflyJARHealthCheckEnricher.BOOTABLE_JAR_ARTIFACT_ID);
+                && JKubeProjectUtil.hasPlugin(getProject(), BOOTABLE_JAR_GROUP_ID, BOOTABLE_JAR_ARTIFACT_ID);
     }
 
     @Override
--- a/jkube-kit/jkube-kit-wildfly-jar/src/test/java/org/eclipse/jkube/wildfly/jar/WildflyJarUtilsTest.java
+++ b/jkube-kit/jkube-kit-wildfly-jar/src/test/java/org/eclipse/jkube/wildfly/jar/WildflyJarUtilsTest.java
@@ -1,0 +1,66 @@
+/**
+ * Copyright (c) 2019 Red Hat, Inc.
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at:
+ *
+ *     https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *   Red Hat, Inc. - initial API and implementation
+ */
+package org.eclipse.jkube.wildfly.jar;
+
+import org.eclipse.jkube.kit.common.Dependency;
+import org.eclipse.jkube.kit.common.JavaProject;
+import org.junit.Before;
+import org.junit.Test;
+
+import java.io.IOException;
+import java.util.Collections;
+import java.util.List;
+import java.util.Properties;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.eclipse.jkube.wildfly.jar.WildflyJarUtils.isStartupEndpointSupported;
+
+public class WildflyJarUtilsTest {
+  private JavaProject project;
+
+  @Before
+  public void setup() throws IOException {
+    project = JavaProject.builder()
+        .properties(new Properties())
+        .build();
+  }
+
+  @Test
+  public void isStartupEndpointSupported_withWildflyJar_Before25_0shouldReturnFalse() {
+    // Given
+    project.setDependencies(createNewJavaProjectDependencyList("24.0.1.Final"));
+    // When
+    boolean result = isStartupEndpointSupported(project);
+    // Then
+    assertThat(result).isFalse();
+  }
+
+  @Test
+  public void isStartupEndpointSupported_withWildflyJar_After25_0shouldReturnTrue() {
+    // Given
+    project.setDependencies(createNewJavaProjectDependencyList("26.1.1.Final"));
+    // When
+    boolean result = isStartupEndpointSupported(project);
+    // Then
+    assertThat(result).isTrue();
+  }
+
+  private List<Dependency> createNewJavaProjectDependencyList(String version) {
+    return Collections.singletonList(Dependency.builder()
+        .groupId("org.wildfly.plugins")
+        .artifactId("wildfly-jar-maven-plugin")
+        .version(version)
+        .build());
+  }
+}
--- a/jkube-kit/jkube-kit-wildfly-jar/src/test/java/org/eclipse/jkube/wildfly/jar/enricher/WildflyJARHealthCheckEnricherTest.java
+++ b/jkube-kit/jkube-kit-wildfly-jar/src/test/java/org/eclipse/jkube/wildfly/jar/enricher/WildflyJARHealthCheckEnricherTest.java
@@ -13,6 +13,21 @@
  */
 package org.eclipse.jkube.wildfly.jar.enricher;
 
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+
+import org.eclipse.jkube.kit.common.Dependency;
+import org.eclipse.jkube.kit.common.JavaProject;
+import org.eclipse.jkube.kit.common.Plugin;
+import org.eclipse.jkube.kit.config.resource.PlatformMode;
+import org.eclipse.jkube.kit.config.resource.ProcessorConfig;
+import org.eclipse.jkube.kit.enricher.api.JKubeEnricherContext;
+import org.eclipse.jkube.kit.enricher.api.model.Configuration;
+
 import com.fasterxml.jackson.core.JsonProcessingException;
 import io.fabric8.kubernetes.api.builder.TypedVisitor;
 import io.fabric8.kubernetes.api.model.ContainerBuilder;
@@ -21,28 +36,14 @@ import io.fabric8.kubernetes.api.model.EnvVarSource;
 import io.fabric8.kubernetes.api.model.KubernetesListBuilder;
 import io.fabric8.kubernetes.api.model.Probe;
 import io.fabric8.kubernetes.api.model.apps.DeploymentBuilder;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Map;
-
 import io.fabric8.kubernetes.client.utils.Serialization;
 import mockit.Expectations;
-import org.eclipse.jkube.kit.enricher.api.JKubeEnricherContext;
 import mockit.Mocked;
-import org.eclipse.jkube.kit.common.JavaProject;
-import org.eclipse.jkube.kit.common.Plugin;
-import org.eclipse.jkube.kit.config.resource.PlatformMode;
-import org.eclipse.jkube.kit.config.resource.ProcessorConfig;
-import org.eclipse.jkube.kit.enricher.api.model.Configuration;
-import org.junit.Assert;
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNotNull;
-import static org.junit.Assert.assertNull;
 import org.junit.Test;
 
+import static org.assertj.core.api.Assertions.assertThat;
+
+@SuppressWarnings({"ResultOfMethodCallIgnored", "unchecked", "unused"})
 public class WildflyJARHealthCheckEnricherTest {
 
     @Mocked
@@ -50,20 +51,17 @@ public class WildflyJARHealthCheckEnricherTest {
 
     private void setupExpectations(JavaProject project, Map<String, Object> bootableJarconfig, Map<String, Map<String, Object>> jkubeConfig) {
         Plugin plugin =
-                Plugin.builder().artifactId(WildflyJARHealthCheckEnricher.BOOTABLE_JAR_ARTIFACT_ID).
-                        groupId(WildflyJARHealthCheckEnricher.BOOTABLE_JAR_GROUP_ID).configuration(bootableJarconfig).build();
+                Plugin.builder().artifactId("wildfly-jar-maven-plugin").
+                        groupId("org.wildfly.plugins").configuration(bootableJarconfig).build();
         List<Plugin> lst = new ArrayList<>();
         lst.add(plugin);
         ProcessorConfig c = new ProcessorConfig(null, null, jkubeConfig);
         new Expectations() {{
-            project.getPlugins();
-            result = lst;
-            context.getProject();
-            result = project;
+            project.getPlugins(); result = lst;
+            context.getProject(); result = project;
             Configuration.ConfigurationBuilder configBuilder = Configuration.builder();
             configBuilder.processorConfig(c);
-            context.getConfiguration();
-            result = configBuilder.build();
+            context.getConfiguration(); result = configBuilder.build();
         }};
     }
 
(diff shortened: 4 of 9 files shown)


jkube-kit/jkube-kit-wildfly-jar/src/main/java/org/eclipse/jkube/wildfly/jar/enricher/WildflyJARHealthCheckEnricher.java
```diff
@@ -112,6 +119,16 @@ private Probe discoverWildflyJARHealthCheck(boolean isReadiness) {
...
 
+            if (isStartup) {
+                int periodSeconds = getPeriodSeconds();
+                scheme = getScheme().toUpperCase();
+                path = getStartupPath();
+                return new ProbeBuilder()
+                        .withNewHttpGet().withNewPort(port).withPath(path).withScheme(scheme).endHttpGet()
+                        .withFailureThreshold(getFailureThreshold())
+                        .withSuccessThreshold(getSuccessThreshold())
+                        .withPeriodSeconds(periodSeconds).build();
```
rohanKanojia, line 130: I see you've added a new periodSeconds config parameter. Why is it used only in case of startup probes?

I see both blocks for ProbeBuilder look the same. Can we change it like this?

```java
            int port = getPort();
            if (port <= 0) {
                return null;
            }
            // scheme must be in upper case in k8s
            String scheme = getScheme().toUpperCase();
            int periodSeconds = getPeriodSeconds();

            return new ProbeBuilder().withNewHttpGet()
                .withNewPort(port).withPath(getProbePath(isReadiness, isStartup)).withScheme(scheme)
                .endHttpGet()
                .withFailureThreshold(getFailureThreshold())
                .withSuccessThreshold(getSuccessThreshold())
                .withInitialDelaySeconds(getInitialDelaySeonds(isReadiness, isStartup))
                .withPeriodSeconds(periodSeconds)
                .build();
```

jkube-kit/jkube-kit-wildfly-jar/src/main/java/org/eclipse/jkube/wildfly/jar/enricher/WildflyJARHealthCheckEnricher.java
```diff
@@ -55,8 +55,10 @@ private enum Config implements Configs.Config {
         SUCCESSTHRESHOLD("successThreshold","1"),
         LIVENESSINITIALDELAY("livenessInitialDelay", "60"),
         READINESSINITIALDELAY("readinessInitialDelay", "10"),
+        PERIODSECONDS("periodSeconds", "10"),
```
rohanKanojia, line 58: Do you think we should also add a startupInitialDelay field?
```
STARTUPINITIALDELAY("startupInitialDelay", "10"),
```

anurag-rajawat: I just refactor periodSeconds to startupInitialDelay

jkube-kit/jkube-kit-wildfly-jar/src/main/java/org/eclipse/jkube/wildfly/jar/enricher/WildflyJARHealthCheckEnricher.java
rohanKanojia: Umm, from what I read in Kubernetes documentation `initialDelaySeconds` and `periodSeconds` are different fields. The former is wait period before performing the first probe and the latter is time interval between performing probes repeatedly.

rohanKanojia: Sorry, I didn't mean to remove `periodSeconds` but use it in the case of all three probes.

anurag-rajawat: Sorry for the misunderstanding and thanks for the clarification.

jkube-kit/jkube-kit-wildfly-jar/src/main/java/org/eclipse/jkube/wildfly/jar/enricher/WildflyJARHealthCheckEnricher.java
anurag-rajawat: According to [Configure Probes](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes), the default value for `periodSeconds` is 10 and if we keep the default value then it's same as `startupInitialDelay` & `readinessInitialDelay` so, do we need to change any of these or keep the default?

rohanKanojia: I think it's okay to keep the default values as provided by kubernetes. If the user needs to change these, an explicit configuration can be provided.

rohanKanojia:
Could you please rebase your branch against master?

anurag-rajawat:
Sure

rohanKanojia:
I think we would need to add startup probes only when the underlying framework supports it (just like we did in case of Quarkus in #1524 ). We would need to check whether Wildfly version is at least `25.0.0.Final`, then only add startup probes

anurag-rajawat:
> Once this gets merged, we can use this method in #1541 and #1492

I was also thinking how to use this, but don't know from which version of wildfly startup probe endpoint was added and you just commented
Thanks

jkube-kit/jkube-kit-wildfly-jar/src/test/java/org/eclipse/jkube/wildfly/jar/WildflyJarUtilsTest.java
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  @Test
+  public void isStartupEndpointSupported_withWildflyJar_Before25_0shouldReturnFalse() {
+    // Given
+    project.setDependencies(createNewJavaProjectDependency("24.0.1.Final"));
+    // When
+    boolean result = isStartupEndpointSupported(project);
+    // Then
+    assertFalse(result);
```
rohanKanojia, line 57: Add AssertJ to project and use assertj assertions here

jkube-kit/jkube-kit-wildfly-jar/src/test/java/org/eclipse/jkube/wildfly/jar/WildflyJarUtilsTest.java
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  @Test
+  public void isStartupEndpointSupported_withWildflyJar_After25_0shouldReturnTrue() {
+    // Given
+    project.setDependencies(createNewJavaProjectDependency("26.1.1.Final"));
+    // When
+    boolean result = isStartupEndpointSupported(project);
+    // Then
+    assertTrue(result);
```
rohanKanojia, line 67: Add AssertJ to project and use assertj assertions here

jkube-kit/jkube-kit-wildfly-jar/src/main/java/org/eclipse/jkube/wildfly/jar/enricher/WildflyJARHealthCheckEnricher.java
```diff
@@ -100,32 +120,32 @@ public void visit(ContainerBuilder containerBuilder) {
...
-            String path = isReadiness ? getReadinessPath() : getLivenessPath();
-
             return new ProbeBuilder()
-                    .withNewHttpGet().withNewPort(port).withPath(path).withScheme(scheme).endHttpGet()
-                    .withFailureThreshold(getFailureThreshold())
-                    .withSuccessThreshold(getSuccessThreshold())
-                    .withInitialDelaySeconds(initialDelay).build();
+                    .withNewHttpGet().withNewPort(asInteger(getConfig(Config.PORT)))
+                    .withPath(path)
+                    .withScheme(getConfig(Config.SCHEME))
```
rohanKanojia, line 132:
```suggestion
                    .withScheme(getConfig(Config.SCHEME).toUpperCase())
```

jkube-kit/jkube-kit-wildfly-jar/src/test/java/org/eclipse/jkube/wildfly/jar/enricher/WildflyJARHealthCheckEnricherTest.java
```diff
@@ -99,22 +107,56 @@ public void testCloudConfiguration(@Mocked final JavaProject project) {
...
+    public void testWithCustomConfigurationComingFromConf_withWildflyJarBefore25_0shouldNotAdd_startupProbe() {
+        wildFlyJarDependencyWithVersion("24.1.1.Final");
+
         Map<String, Object> jarConfig = new HashMap<>();
         jarConfig.put("cloud", null);
         final Map<String, Map<String, Object>> config = createFakeConfig("{\"readinessPath\":\"/foo/ready\","
                 + "\"livenessPath\":\"/foo/live\","
                 + "\"port\":\"1234\","
-                + "\"scheme\":\"https\","
+                + "\"scheme\":\"HTTPS\","
```
rohanKanojia, line 153: You shouldn't be changing existing test expectations. It was failing because we missed `toUpperCase` while parsing scheme

jkube-kit/jkube-kit-wildfly-jar/src/test/java/org/eclipse/jkube/wildfly/jar/enricher/WildflyJARHealthCheckEnricherTest.java
```diff
@@ -78,71 +82,170 @@ private void setupExpectations(Map<String, Map<String, Object>> jkubeConfig) {
...
-        assertEquals(17, probe.getFailureThreshold().intValue());
+        assertThat(probe).isNotNull();
+        assertThat(probe.getHttpGet().getHost()).isNull();
+        assertThat(probe.getHttpGet().getScheme()).isEqualTo("HTTPS");
+        assertThat(probe.getHttpGet().getPort().getIntVal().intValue()).isEqualTo(1234);
+        assertThat(probe.getHttpGet().getPath()).isEqualTo("/foo/live");
+        assertThat(probe.getInitialDelaySeconds().intValue()).isEqualTo(99);
+        assertThat(probe.getPeriodSeconds().intValue()).isEqualTo(10);
+        assertThat(probe.getSuccessThreshold().intValue()).isEqualTo(27);
+        assertThat(probe.getFailureThreshold().intValue()).isEqualTo(17);
```
rohanKanojia, line 222: Try to use chained assertions using assertj when asserting specific object. For example, here you can do something like this:

```java
        assertThat(probe)
            .isNotNull()
            .hasFieldOrPropertyWithValue("httpGet.host", null)
            .hasFieldOrPropertyWithValue("httpGet.scheme", "HTTPS")
            .hasFieldOrPropertyWithValue("httpGet.port", new IntOrString(1234))
            .hasFieldOrPropertyWithValue("httpGet.path", "/foo/live")
            .hasFieldOrPropertyWithValue("initialDelaySeconds", 99)
            .hasFieldOrPropertyWithValue("periodSeconds",  10)
            .hasFieldOrPropertyWithValue("successThreshold", 27)
            .hasFieldOrPropertyWithValue("failureThreshold", 17);
```

jkube-kit/jkube-kit-wildfly-jar/src/test/java/org/eclipse/jkube/wildfly/jar/enricher/WildflyJARHealthCheckEnricherTest.java
```diff
@@ -78,71 +82,170 @@ private void setupExpectations(Map<String, Map<String, Object>> jkubeConfig) {
...
-        assertEquals(17, probe.getFailureThreshold().intValue());
+        assertThat(probe).isNotNull();                                    
+        assertThat(probe.getHttpGet().getHost()).isNull();                
+        assertThat(probe.getHttpGet().getScheme()).isEqualTo("HTTPS");
+        assertThat(probe.getHttpGet().getPort().getIntVal().intValue()).isEqualTo(1234);
+        assertThat(probe.getHttpGet().getPath()).isEqualTo("/foo/ready");
+        assertThat(probe.getInitialDelaySeconds().intValue()).isEqualTo(77);
+        assertThat(probe.getPeriodSeconds().intValue()).isEqualTo(30);
+        assertThat(probe.getSuccessThreshold().intValue()).isEqualTo(27);
+        assertThat(probe.getFailureThreshold().intValue()).isEqualTo(17);
```
rohanKanojia, line 233: chain these assertions together like I suggested above

anurag-rajawat: Sure

jkube-kit/jkube-kit-wildfly-jar/src/test/java/org/eclipse/jkube/wildfly/jar/enricher/WildflyJARHealthCheckEnricherTest.java
```diff
@@ -78,71 +82,170 @@ private void setupExpectations(Map<String, Map<String, Object>> jkubeConfig) {
...
+        Probe probe = enricher.getLivenessProbe();
+        assertThat(probe).isNotNull();
+        assertThat(probe.getHttpGet().getHost()).isNull();
+        assertThat(probe.getHttpGet().getScheme()).isEqualTo("HTTPS");
+        assertThat(probe.getHttpGet().getPort().getIntVal().intValue()).isEqualTo(1234);
+        assertThat(probe.getHttpGet().getPath()).isEqualTo("/foo/live");
+        assertThat(probe.getInitialDelaySeconds().intValue()).isEqualTo(99);
+        assertThat(probe.getPeriodSeconds().intValue()).isEqualTo(10);
+        assertThat(probe.getSuccessThreshold().intValue()).isEqualTo(27);
+        assertThat(probe.getFailureThreshold().intValue()).isEqualTo(17);
```
rohanKanojia, line 174: I see there are a lot of places where we're asserting `Probe` contents. Do you think we should create a new method `assertProbe` that would receive probe to verify and expected values?

anurag-rajawat: It would be nice as it was done in  [Quarkus Test](https://github.com/eclipse/jkube/blob/0e8269d1a290a3542a1144a40310e9c39f69d566/jkube-kit/jkube-kit-quarkus/src/test/java/org/eclipse/jkube/quarkus/enricher/QuarkusHealthCheckEnricherTest.java#L285)

anurag-rajawat: Should we do something like this?
```java
    private void assertProbe(Probe probe, String scheme, int port, String path, int initialDelaySeconds,
                             int periodSeconds, int successThreshold, int failureThreshold) {
        assertThat(probe)
                .isNotNull()
                .hasFieldOrPropertyWithValue("httpGet.host", null)
                .hasFieldOrPropertyWithValue("httpGet.scheme", scheme)
                .hasFieldOrPropertyWithValue("httpGet.port", new IntOrString(port))
                .hasFieldOrPropertyWithValue("httpGet.path", path)
                .hasFieldOrPropertyWithValue("initialDelaySeconds", initialDelaySeconds)
                .hasFieldOrPropertyWithValue("periodSeconds",  periodSeconds)
                .hasFieldOrPropertyWithValue("successThreshold", successThreshold)
                .hasFieldOrPropertyWithValue("failureThreshold", failureThreshold);
    }
```

rohanKanojia: looks good, but I would move httpGet related assertions to another method `assertHttpGet` and use that here. I see some places where we only check httpGet. This method could be reused there as well.

rohanKanojia:
Nice work!

jkube-kit/jkube-kit-wildfly-jar/src/main/java/org/eclipse/jkube/wildfly/jar/enricher/WildflyJARHealthCheckEnricher.java
```diff
@@ -67,12 +79,20 @@ private enum Config implements Configs.Config {
 
     @Override
     protected Probe getReadinessProbe() {
-        return discoverWildflyJARHealthCheck(true);
+        return discoverWildflyJARHealthCheck(getConfig(Config.READINESS_PATH), asInt(getConfig(Config.READINESS_INITIAL_DELAY)), asInt(getConfig(Config.FAILURE_THRESHOLD)), asInt(getConfig(Config.SUCCESS_THRESHOLD)), asInt(getConfig(Config.READINESS_PERIOD_SECONDS)));
```
manusa, line 82: Considering that in the end we're passing the values extracted from configurations by following the same pattern in all of the `discoverWildflyJARHealthCheck` method calls, why not just pass the Configuration reference?

```suggestion
        return discoverWildflyJARHealthCheck(Config.READINESS_PATH, Config.READINESS_INITIAL_DELAY, Config.FAILURE_THRESHOLD, Config.SUCCESS_THRESHOLD, Config.READINESS_PERIOD_SECONDS);
```

jkube-kit/jkube-kit-wildfly-jar/src/main/java/org/eclipse/jkube/wildfly/jar/enricher/WildflyJARHealthCheckEnricher.java
```diff
@@ -100,32 +120,32 @@ public void visit(ContainerBuilder containerBuilder) {
         super.create(platformMode, builder);
     }
 
-    private Probe discoverWildflyJARHealthCheck(boolean isReadiness) {
-
+    private Probe discoverWildflyJARHealthCheck(String path, int initialDelay, int failureThreshold, int successThreshold, int periodSeconds) {
```
manusa, line 123: See comment about passing Configuration(s) instead of resolved values.

jkube-kit/jkube-kit-wildfly-jar/src/main/java/org/eclipse/jkube/wildfly/jar/enricher/WildflyJARHealthCheckEnricher.java
```diff
@@ -100,32 +120,32 @@ public void visit(ContainerBuilder containerBuilder) {
...
+                    .withInitialDelaySeconds(initialDelay)
+                    .withPeriodSeconds(periodSeconds)
+                    .build();
         }
         return null;
     }
 
     private boolean isAvailable() {
-        if (isProbeEnforced()) {
+        boolean isProbeEnforced = asBoolean(getConfig(Config.ENFORCE_PROBES));
```
manusa, line 144: is this variable used elsewhere? why not just evaluate in the if clause?

jkube-kit/jkube-kit-wildfly-jar/src/main/java/org/eclipse/jkube/wildfly/jar/enricher/WildflyJARHealthCheckEnricher.java
```diff
@@ -134,39 +154,16 @@ private boolean isAvailable() {
...
-        return Configs.asInteger(getConfig(Config.SUCCESSTHRESHOLD));
-    }
-
-    protected String getScheme() {
-        return Configs.asString(getConfig(Config.SCHEME));
-    }
-
     protected int getPort() {
-        return Configs.asInt(getConfig(Config.PORT));
+        return asInt(getConfig(Config.PORT));
```
manusa, line 158: Is this method still used?? I see we replaced the call to `getPort()` in the `discoverWildflyJARHealthCheck`, this method is probably useless now.

jkube-kit/jkube-kit-wildfly-jar/src/main/java/org/eclipse/jkube/wildfly/jar/enricher/WildflyJARHealthCheckEnricher.java
```diff
@@ -134,39 +154,16 @@ private boolean isAvailable() {
...
     }
 
     protected String getLivenessPath() {
```
manusa, line 161: Not used anymore.

(Tests should be improved to test the actual Enricher behavior)

jkube-kit/jkube-kit-wildfly-jar/src/main/java/org/eclipse/jkube/wildfly/jar/enricher/WildflyJARHealthCheckEnricher.java
```diff
@@ -134,39 +154,16 @@ private boolean isAvailable() {
...
-        return Configs.asString(getConfig(Config.LIVENESSPATH));
-    }
-
-    protected int getLivenessInitialDelay() {
-        return Configs.asInt(getConfig(Config.LIVENESSINITIALDELAY));
+        return Configs.asString(getConfig(Config.LIVENESS_PATH));
     }
 
     protected String getReadinessPath() {
```
manusa, line 165: Not used anymore.

(Tests should be improved to test the actual Enricher behavior)

jkube-kit/jkube-kit-wildfly-jar/src/test/java/org/eclipse/jkube/wildfly/jar/WildflyJarUtilsTest.java
```diff
@@ -0,0 +1,70 @@
...
+    // When
+    boolean result = isStartupEndpointSupported(project);
+    // Then
+    assertThat(result).isTrue();
+  }
+
+  private List<Dependency> createNewJavaProjectDependencyList(String version) {
+    return Collections.singletonList(Dependency.builder()
+        .groupId(BOOTABLE_JAR_GROUP_ID)
+        .artifactId(BOOTABLE_JAR_ARTIFACT_ID)
```
manusa, line 66: Please use string literals and not constants here

jkube-kit/jkube-kit-wildfly-jar/src/test/java/org/eclipse/jkube/wildfly/jar/enricher/WildflyJARHealthCheckEnricherTest.java
```diff
@@ -13,45 +13,52 @@
...
+import static org.eclipse.jkube.wildfly.jar.WildflyJarUtils.BOOTABLE_JAR_ARTIFACT_ID;
+import static org.eclipse.jkube.wildfly.jar.WildflyJarUtils.BOOTABLE_JAR_GROUP_ID;
+
+@SuppressWarnings({"ResultOfMethodCallIgnored", "unchecked", "unused"})
 public class WildflyJARHealthCheckEnricherTest {
 
     @Mocked
     protected JKubeEnricherContext context;
+    @Mocked
+    private JavaProject project;
```
manusa, line 56: Maybe use a real JavaProject instance instead of a mocked one.

jkube-kit/jkube-kit-wildfly-jar/src/test/java/org/eclipse/jkube/wildfly/jar/enricher/WildflyJARHealthCheckEnricherTest.java
```diff
@@ -13,45 +13,52 @@
...
 
     private void setupExpectations(JavaProject project, Map<String, Object> bootableJarconfig, Map<String, Map<String, Object>> jkubeConfig) {
         Plugin plugin =
-                Plugin.builder().artifactId(WildflyJARHealthCheckEnricher.BOOTABLE_JAR_ARTIFACT_ID).
-                        groupId(WildflyJARHealthCheckEnricher.BOOTABLE_JAR_GROUP_ID).configuration(bootableJarconfig).build();
+                Plugin.builder().artifactId(BOOTABLE_JAR_ARTIFACT_ID).
```
manusa, line 60: Please use string literals and not constants

jkube-kit/jkube-kit-wildfly-jar/src/test/java/org/eclipse/jkube/wildfly/jar/enricher/WildflyJARHealthCheckEnricherTest.java
```diff
@@ -78,71 +85,120 @@ private void setupExpectations(Map<String, Map<String, Object>> jkubeConfig) {
...
     @Test
-    public void testDefaultConfiguration(@Mocked final JavaProject project) {
+    public void testDefaultConfiguration() {
         setupExpectations(project, Collections.emptyMap(), Collections.emptyMap());
         WildflyJARHealthCheckEnricher enricher = new WildflyJARHealthCheckEnricher(context);
         Probe probe = enricher.getLivenessProbe();
-        assertNull(probe);
+        assertThat(probe).isNull();
+        probe = enricher.getReadinessProbe();
+        assertThat(probe).isNull();
```
manusa, line 94: Don't reuse the variable:

```java
assertThat(enricher)
  .returns(null, e -> e.getLivenessProbe())
  .returns(null, e -> e.getReadinessProbe());
```

jkube-kit/jkube-kit-wildfly-jar/src/test/java/org/eclipse/jkube/wildfly/jar/enricher/WildflyJARHealthCheckEnricherTest.java
```diff
@@ -78,71 +85,120 @@ private void setupExpectations(Map<String, Map<String, Object>> jkubeConfig) {
...
+    @Test
+    public void testCloudConfiguration_withWildflyJarBefore25_0shouldNotAdd_startupProbe() {
+        wildFlyJarDependencyWithVersion("24.1.1.Final");
+        
+        Map<String, Object> config = new HashMap<>();
+        config.put("cloud", null);
+        setupExpectations(project, config, Collections.emptyMap());
+        WildflyJARHealthCheckEnricher enricher = new WildflyJARHealthCheckEnricher(context);
+
+        Probe probe = enricher.getLivenessProbe();
```
manusa, line 106: Please, don't reuse the local variable

jkube-kit/jkube-kit-wildfly-jar/src/test/java/org/eclipse/jkube/wildfly/jar/enricher/WildflyJARHealthCheckEnricherTest.java
```diff
@@ -78,71 +85,120 @@ private void setupExpectations(Map<String, Map<String, Object>> jkubeConfig) {
...
+    public void testCloudConfiguration_withWildflyJarAfter25_0shouldAdd_startupProbe() {
+        // Wildfly version must be at least 25.0.0.Final to add startup probe
+        wildFlyJarDependencyWithVersion("26.1.1.Final");
+
         Map<String, Object> config = new HashMap<>();
         config.put("cloud", null);
         setupExpectations(project, config, Collections.emptyMap());
         WildflyJARHealthCheckEnricher enricher = new WildflyJARHealthCheckEnricher(context);
+
         Probe probe = enricher.getLivenessProbe();
```
manusa, line 126: Please, don't reuse the local variable

jkube-kit/jkube-kit-wildfly-jar/src/test/java/org/eclipse/jkube/wildfly/jar/enricher/WildflyJARHealthCheckEnricherTest.java
```diff
@@ -186,18 +262,18 @@ public void testEnforceFalseHealth() {
...
     @Test
     public void configureWildFlyJarHealthPort() {
         final WildflyJARHealthCheckEnricher enricher = new WildflyJARHealthCheckEnricher(context);
         final int port = enricher.getPort();
-        Assert.assertEquals(9990, port);
-        Assert.assertEquals("/health/live", enricher.getLivenessPath());
-        Assert.assertEquals("/health/ready", enricher.getReadinessPath());
+        assertThat(port).isEqualTo(9990);
+        assertThat(enricher.getLivenessPath()).isEqualTo("/health/live");
+        assertThat(enricher.getReadinessPath()).isEqualTo("/health/ready");
```
manusa, line 276: Please combine the assertions

jkube-kit/jkube-kit-wildfly-jar/src/test/java/org/eclipse/jkube/wildfly/jar/enricher/WildflyJARHealthCheckEnricherTest.java
```diff
@@ -224,15 +300,26 @@ public void visit(ContainerBuilder containerBuilder) {
...
-        Assert.assertEquals("metadata.name", src.getFieldRef().getFieldPath());
+        assertThat(src.getFieldRef().getFieldPath()).isEqualTo("metadata.name");
+    }
+
+    private void wildFlyJarDependencyWithVersion(String wildflyJarVersion) {
+        new Expectations() {{
+            project.getDependencies();
+            result = Collections.singletonList(Dependency.builder()
+                .groupId(BOOTABLE_JAR_GROUP_ID)
+                .artifactId(BOOTABLE_JAR_ARTIFACT_ID)
```
manusa, line 319: Please use literals instead of constants here.

jkube-kit/jkube-kit-wildfly-jar/src/test/java/org/eclipse/jkube/wildfly/jar/generator/WildflyJARGeneratorTest.java
```diff
@@ -157,84 +153,77 @@ public void slimServerNoDir(@Mocked final JavaProject project) throws Exception
...
+                .first().isEqualTo("-Djava.net.preferIPv4Stack=true");
+        assertThat(extraOptions.get(1)).isEqualTo("-Dmaven.repo.local=/deployments/myrepo");
     }
     
     private GeneratorContext contextForSlimServer(JavaProject project, Map<String, Object> bootableJarconfig, Path dir) {
         Plugin plugin
-                = Plugin.builder().artifactId(WildflyJARHealthCheckEnricher.BOOTABLE_JAR_ARTIFACT_ID).
-                        groupId(WildflyJARHealthCheckEnricher.BOOTABLE_JAR_GROUP_ID).configuration(bootableJarconfig).build();
+                = Plugin.builder().artifactId(BOOTABLE_JAR_ARTIFACT_ID).
+                        groupId(BOOTABLE_JAR_GROUP_ID).configuration(bootableJarconfig).build();
```
manusa, line 226: Please use literals instead of constants here

manusa (changes requested):
Looking good, but there are a few things to be changed

jkube-kit/jkube-kit-wildfly-jar/src/main/java/org/eclipse/jkube/wildfly/jar/enricher/WildflyJARHealthCheckEnricher.java
anurag-rajawat: This and the other two methods `getLivenessPath()` and `getReadinessPath()` are only required for this test
https://github.com/eclipse/jkube/blob/e643a0a28ffe520b3122a42cf01cf0792d4fd1ba/jkube-kit/jkube-kit-wildfly-jar/src/test/java/org/eclipse/jkube/wildfly/jar/enricher/WildflyJARHealthCheckEnricherTest.java#L195-L201

anurag-rajawat: Do we really need this test? If yes then tell me how to test the actual Enricher behavior after removing these three methods.

jkube-kit/jkube-kit-wildfly-jar/src/main/java/org/eclipse/jkube/wildfly/jar/enricher/WildflyJARHealthCheckEnricher.java
manusa: This wasn't fixed
We still have a local variable declaration that is not needed. Why aren't we just consuming its instantiation value?

```java
        if (asBoolean(getConfig(Config.ENFORCE_PROBES))) {
            return true;
        }

```

jkube-kit/jkube-kit-wildfly-jar/src/main/java/org/eclipse/jkube/wildfly/jar/enricher/WildflyJARHealthCheckEnricher.java
manusa: Are you going to address this? Is there something else I can add or elaborate on to clarify what I mean?

jkube-kit/jkube-kit-wildfly-jar/src/main/java/org/eclipse/jkube/wildfly/jar/enricher/WildflyJARHealthCheckEnricher.java
manusa: Testing an unused method seems like something unnecessary.
We should refactor the test to actually verify something. IF we changed the implementation of the calling method to evaluate the ports, then refactor the test to verify that method.

> Do we really need this test? 

That's part of your assignment. You should try to evaluate what the class is actually doing and if the tests are verifying this behavior.

jkube-kit/jkube-kit-wildfly-jar/src/test/java/org/eclipse/jkube/wildfly/jar/WildflyJarUtilsTest.java
manusa: I can still see the constants here

manusa: If you're using the resolution feature to track that you locally changed things, maybe use an emoji instead.
It makes it very hard for reviewers when you mark things as resolved but things haven't changed in the repo. Personally, I prefer to resolve the conversations so I can recall what I requested changes for (instead of expanding all of the conversations).

manusa (changes requested):
I see there are still many pending changes.
What's your status on this?

jkube-kit/jkube-kit-wildfly-jar/src/main/java/org/eclipse/jkube/wildfly/jar/enricher/WildflyJARHealthCheckEnricher.java
anurag-rajawat: > Are you going to address this?

Yes


> Is there something else I can add or elaborate on to clarify what I mean?

No

jkube-kit/jkube-kit-wildfly-jar/src/main/java/org/eclipse/jkube/wildfly/jar/enricher/WildflyJARHealthCheckEnricher.java
```diff
@@ -100,73 +141,17 @@ public void visit(ContainerBuilder containerBuilder) {
...
-    }
-
     private boolean isAvailable() {
-        if (isProbeEnforced()) {
+        if (asBoolean(getConfig(Config.ENFORCE_PROBES))) {
             return true;
         }
         JavaProject project = getContext().getProject();
-        Plugin plugin = JKubeProjectUtil.getPlugin(project, BOOTABLE_JAR_GROUP_ID, BOOTABLE_JAR_ARTIFACT_ID);
+        Plugin plugin = JKubeProjectUtil.getPlugin(project, "org.wildfly.plugins", "wildfly-jar-maven-plugin");
```
manusa, line 149: I'm not sure if you interpreted this from a comment in one of the tests.

**Production code** should be using the constants, that's not a problem.

However, the test code SHOULD NOT use the constants, especially if they are the same as the ones in the production code. The reason is simple, a change in the constant, will not make the test fail even though we've completely changed the specification constraint. Considering a test is used to verify a specification or a defined behavior, we should use either a literal, or if needed a variable defined in the same test (however, this might make the test less readable).

jkube-kit/jkube-kit-wildfly-jar/src/main/java/org/eclipse/jkube/wildfly/jar/generator/WildflyJARGenerator.java
```diff
@@ -42,7 +40,7 @@ public class WildflyJARGenerator extends JavaExecGenerator {
     public WildflyJARGenerator(GeneratorContext context) {
         super(context, "wildfly-jar");
         JavaProject project = context.getProject();
-        Plugin plugin = JKubeProjectUtil.getPlugin(project, BOOTABLE_JAR_GROUP_ID, BOOTABLE_JAR_ARTIFACT_ID);
+        Plugin plugin = JKubeProjectUtil.getPlugin(project, "org.wildfly.plugins", "wildfly-jar-maven-plugin");
```
manusa, line 43: I'm not sure if you interpreted this from a comment in one of the tests.

**Production code** should be using the constants, that's not a problem.

However, the test code SHOULD NOT use the constants, especially if they are the same as the ones in the production code. The reason is simple, a change in the constant, will not make the test fail even though we've completely changed the specification constraint. Considering a test is used to verify a specification or a defined behavior, we should use either a literal, or if needed a variable defined in the same test (however, this might make the test less readable).

jkube-kit/jkube-kit-wildfly-jar/src/main/java/org/eclipse/jkube/wildfly/jar/generator/WildflyJARGenerator.java
```diff
@@ -57,7 +55,7 @@ public WildflyJARGenerator(GeneratorContext context) {
     public boolean isApplicable(List<ImageConfiguration> configs) {
         return shouldAddGeneratedImageConfiguration(configs)
                 && JKubeProjectUtil.hasPlugin(getProject(),
-                        WildflyJARHealthCheckEnricher.BOOTABLE_JAR_GROUP_ID, WildflyJARHealthCheckEnricher.BOOTABLE_JAR_ARTIFACT_ID);
+                        "org.wildfly.plugins", "wildfly-jar-maven-plugin");
```
manusa, line 58: I'm not sure if you interpreted this from a comment in one of the tests.

**Production code** should be using the constants, that's not a problem.

However, the test code SHOULD NOT use the constants, especially if they are the same as the ones in the production code. The reason is simple, a change in the constant, will not make the test fail even though we've completely changed the specification constraint. Considering a test is used to verify a specification or a defined behavior, we should use either a literal, or if needed a variable defined in the same test (however, this might make the test less readable).

jkube-kit/jkube-kit-wildfly-jar/src/test/java/org/eclipse/jkube/wildfly/jar/enricher/WildflyJARHealthCheckEnricherTest.java
```diff
@@ -81,64 +81,120 @@ private void setupExpectations(Map<String, Map<String, Object>> jkubeConfig) {
...
+                + "\"startupInitialDelay\":\"10\","
+                + "\"startupFailureThreshold\":\"3\","
+                + "\"startupSuccessThreshold\":\"1\","
+                + "\"startupPeriodSeconds\":\"10\""
+                + "}");
+        setupExpectations(project, jarConfig, config);
+
+        final WildflyJARHealthCheckEnricher enricher = new WildflyJARHealthCheckEnricher(context);
+        final Probe livenessProbe = enricher.getLivenessProbe();
+        assertProbesAdded(livenessProbe, "HTTPS", "/foo/live", 1234, Arrays.asList(99, 3, 1, 10));
```
manusa, line 191: This is not very readable, it's hard to say what the `99, 3, 1, 10` values mean.

Especially considering that these values are repeated:
A simple replacement of the Config.READINESS_SUCCESS_THRESHOLD with the Config.STARTUP_SUCCESS_THRESHOLD in the `getReadinessProbe()` wouldn't make this test fail.

Also, some of these values are just the default, value, so a change in the first argument of the Config enum for READINESS_SUCCESS_THRESHOLD wouldn't make the test fail either.

manusa (changes requested):
We're almost there.

jkube-kit/jkube-kit-wildfly-jar/src/main/java/org/eclipse/jkube/wildfly/jar/enricher/WildflyJARHealthCheckEnricher.java
anurag-rajawat: Got it

jkube-kit/jkube-kit-wildfly-jar/src/main/java/org/eclipse/jkube/wildfly/jar/generator/WildflyJARGenerator.java
anurag-rajawat: Got it

jkube-kit/jkube-kit-wildfly-jar/src/main/java/org/eclipse/jkube/wildfly/jar/generator/WildflyJARGenerator.java
anurag-rajawat: Got it

anurag-rajawat:
@manusa please review this PR

manusa (approved):
LGTM, thx!

anurag-rajawat:
@manusa sorry for the noise and thanks a lot for helping me.

PR #1508, opened 2022-05-19
chore : Remove vertx-infinispan dependency from jkube-kit

rohanKanojia:
## Description

Infinispan dependency was only getting used in jkube-kit-vertx under
test scope only. We can easily work around this by mocking static method
as per our requirement

## Type of change

## Checklist

### DIFF
--- a/jkube-kit/jkube-kit-vertx/src/test/java/org/eclipse/jkube/vertx/generator/VertxGeneratorTest.java
+++ b/jkube-kit/jkube-kit-vertx/src/test/java/org/eclipse/jkube/vertx/generator/VertxGeneratorTest.java
@@ -23,35 +23,32 @@ import org.eclipse.jkube.kit.common.JavaProject;
 import org.eclipse.jkube.kit.common.Dependency;
 import org.eclipse.jkube.kit.common.KitLogger;
 import org.eclipse.jkube.generator.api.GeneratorContext;
-import mockit.Expectations;
-import mockit.Injectable;
-import mockit.Mocked;
+import org.eclipse.jkube.kit.common.util.JKubeProjectUtil;
 import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.TemporaryFolder;
+import org.mockito.MockedStatic;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Answers.RETURNS_DEEP_STUBS;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.mockStatic;
+import static org.mockito.Mockito.when;
 
 /**
  * @author <a href="http://escoffier.me">Clement Escoffier</a>
  */
-@SuppressWarnings("ResultOfMethodCallIgnored")
 public class VertxGeneratorTest {
-
-  @Injectable
-  private KitLogger logger;
 
   @Rule
   public TemporaryFolder folder = new TemporaryFolder();
 
-  @Mocked
   private JavaProject project;
 
   private GeneratorContext context;
   private Dependency dropwizard;
   private Dependency core;
-  private Dependency infinispan;
 
   @Before
   public void init() throws IOException {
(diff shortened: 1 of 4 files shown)


manusa:
This looks good, but I don't get why the real dependency was required previously. There were also mocks to declare that project dependency.

rohanKanojia:
I think this is because of this line:

https://github.com/eclipse/jkube/blob/7e849a7dd04466ea10684e0c493c3e5dcd5db1ea/jkube-kit/jkube-kit-vertx/src/main/java/org/eclipse/jkube/vertx/generator/VertxGenerator.java#L91

Here we're checking whether `META-INF/services/io.vertx.core.spi.cluster.ClusterManager` resource is in classpath or not. This resource is present in vertx-infinispan dependency. Now that I've mocked `JKubeProjectUtil.hasResource` this shouldn't be checked.

jkube-kit/jkube-kit-vertx/src/test/java/org/eclipse/jkube/vertx/generator/VertxGeneratorTest.java
```diff
@@ -101,20 +90,23 @@ public void testWithMetrics() {
...
-                      "-Djava.net.preferIPv4Stack=true")
-              .containsEntry("JAVA_ARGS", "-cluster");
+      try (MockedStatic<JKubeProjectUtil> mockedJKubeProjectUtil = mockStatic(JKubeProjectUtil.class)) {
+        // Given
+        mockedJKubeProjectUtil
+            .when(() -> JKubeProjectUtil.hasResource(project, "META-INF/services/io.vertx.core.spi.cluster.ClusterManager"))
+            .thenReturn(true);
+        mockedJKubeProjectUtil
+            .when(() -> JKubeProjectUtil.hasDependency(project, "io.vertx", "vertx-infinispan"))
+            .thenReturn(true);
```
manusa, line 100: Considering this: https://github.com/eclipse/jkube/pull/1508#issuecomment-1134224352

Then this line should be changed to what was before, to make all test cases consistent:
```java
when(project.getDependencies()).thenReturn(Arrays.asList(infinispan, core));
```

manusa (changes requested):

jkube-kit/jkube-kit-vertx/src/test/java/org/eclipse/jkube/vertx/generator/VertxGeneratorTest.java
rohanKanojia: I tried reverting it but I started getting this error:

```
java.lang.AssertionError: 
Expecting map:
 <{"JAVA_ARGS"="-cluster", "JAVA_OPTIONS"="-Dvertx.cacheDirBase=/tmp/vertx-cache -Dvertx.disableDnsResolver=true"}>
to contain:
 <[MapEntry[key="JAVA_OPTIONS", value="-Dvertx.cacheDirBase=/tmp/vertx-cache -Dvertx.disableDnsResolver=true -Djava.net.preferIPv4Stack=true"]]>
but could not find the following map entries:
 <[MapEntry[key="JAVA_OPTIONS", value="-Dvertx.cacheDirBase=/tmp/vertx-cache -Dvertx.disableDnsResolver=true -Djava.net.preferIPv4Stack=true"]]>
```
It looks like control is not going in this if block as expected by test:
https://github.com/eclipse/jkube/blob/7e849a7dd04466ea10684e0c493c3e5dcd5db1ea/jkube-kit/jkube-kit-vertx/src/main/java/org/eclipse/jkube/vertx/generator/VertxGenerator.java#L74


I think it's because now we're mocking static methods in JKubeProjectUtil, real `hasDependency`is not invoked.

manusa: OK, we can keep it this way.
We could tune the mock to call the real method, but then the test preparation would be even more verbose.

manusa (approved):

Issue #443, opened 2020-10-20
Enricher for MicroProfile Health implementations


manusa:
### Description
Provide HealthCheck enricher for [MicroProfile Health](https://github.com/eclipse/microprofile-health) implementations:
- [ ] [SmallRye Health](https://github.com/smallrye/smallrye-health) + Quarkified version
- [ ] ...

#### See also:
- https://github.com/eclipse/microprofile-health/blob/master/spec/src/main/asciidoc/protocol-wireformat.adoc

### Info

* Eclipse JKube version :
* Maven version (`mvn -v`) :
```

```
* Kubernetes / Red Hat OpenShift setup and version :

* If it's a bug, how to reproduce :

* If it's a feature request, what is your use case :

* Sample Reproducer Project : *[GitHub Clone URL]*

manusa:
Internal reference: [protean-dev] Philosophy around healthchecks

sunix:
1. This issue for the implementation
2. IT for each specific frameworks: https://github.com/eclipse/jkube/issues/1473 etc...

rohanKanojia:
I think we have health check implementations for the following Microprofile implementations:

- OpenLiberty: https://github.com/eclipse/jkube/pull/1541
- Quarkus: [QuarkusHealthCheckEnricher](https://github.com/eclipse/jkube/blob/master/jkube-kit/jkube-kit-quarkus/src/main/java/org/eclipse/jkube/quarkus/enricher/QuarkusHealthCheckEnricher.java)
- Wildfly: [WildflyJARHealthCheckEnricher](https://github.com/eclipse/jkube/blob/master/jkube-kit/jkube-kit-wildfly-jar/src/main/java/org/eclipse/jkube/wildfly/jar/enricher/WildflyJARHealthCheckEnricher.java)
- ThorntailV2: [ThorntailV2HealthCheckEnricher](https://github.com/eclipse/jkube/blob/master/jkube-kit/jkube-kit-thorntail-v2/src/main/java/org/eclipse/jkube/thorntail/v2/enricher/ThorntailV2HealthCheckEnricher.java)

In [Microprofile Starter](https://start.microprofile.io/), I also see some other Microprofile runtimes for old versions of Microprofile (`3.3`):
- [Payara Micro](https://www.payara.fish/products/payara-micro/)
- [KumuluzEE](https://ee.kumuluz.com/)
- [Helidon](https://helidon.io/)

[...]

PR #1709, opened 2022-08-19
feat (jkube-kit) : Add health check enricher for SmallRye Health (#443)

rohanKanojia:
## Description
Related to #443

+ Add SmallRyeHealthCheckEnricher to add readiness, liveness and startup
probes if `io.smallrye:smallrye-health` is present in project
dependencies.
+ Moved common logic between SmallRyeHealthCheckEnricher and
      OpenLibertyHealthCheckEnricher to `jkube-kit-microprofile` module

## Type of change

## Checklist

### DIFF
--- a/jkube-kit/jkube-kit-microprofile/src/main/java/org/eclipse/jkube/microprofile/MicroprofileHealthUtil.java
+++ b/jkube-kit/jkube-kit-microprofile/src/main/java/org/eclipse/jkube/microprofile/MicroprofileHealthUtil.java
@@ -11,15 +11,17 @@
  * Contributors:
  *   Red Hat, Inc. - initial API and implementation
  */
-package org.eclipse.jkube.kit.common.util;
+package org.eclipse.jkube.microprofile;
 
 import org.eclipse.jkube.kit.common.Dependency;
 import org.eclipse.jkube.kit.common.JavaProject;
+import org.eclipse.jkube.kit.common.util.JKubeProjectUtil;
 
 import static org.eclipse.jkube.kit.common.util.SemanticVersionUtil.isVersionAtLeast;
 
 public class MicroprofileHealthUtil {
-  private static final String MICROPROFILE_HEALTH_DEPENDENCY = "microprofile-health";
+  private static final String MICROPROFILE_HEALTH_DEPENDENCY = "microprofile-health-api";
+  private static final String MICROPROFILE_HEALTH_GROUP = "org.eclipse.microprofile.health";
   private static final String MICROPROFILE_DEPENDENCY = "microprofile";
   private static final String MICROPROFILE_GROUP = "org.eclipse.microprofile";
   private static final int MICROPROFILE_HEALTH_SUPPORTED_VERSION_MAJOR = 3;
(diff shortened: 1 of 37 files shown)


jkube-kit/jkube-kit-microprofile/src/main/java/org/eclipse/jkube/microprofile/MicroprofileHealthUtil.java
```diff
@@ -31,31 +33,31 @@ public class MicroprofileHealthUtil {
...
       return false;
     }
 
     return isVersionAtLeast(MICROPROFILE_HEALTH_SUPPORTED_VERSION_MAJOR, MICROPROFILE_HEALTH_SUPPORTED_VERSION_MINOR, microProfileHealthVersion);
   }
 
-  private static String getMicroProfileVersionFromArtifactId(JavaProject javaProject, String artifactId) {
-    Dependency microProfileDep = JKubeProjectUtil.getDependency(javaProject, MICROPROFILE_GROUP, artifactId);
+  private static String getMicroProfileVersionFromArtifactId(JavaProject javaProject, String groupId, String artifactId) {
+    Dependency microProfileDep = JKubeProjectUtil.getTransitiveDependency(javaProject, groupId, artifactId);
```
manusa, line 57: Just to double check, this does work with Gradle right? 

(The question is redundant since I see the IT, but it's best if we keep your confirmation to this comment as a record in case there are doubts in the future)

manusa (approved):

manusa:
I'd expect this Enricher also affecting Quarkus, but I didn't see any changes in that module.
Are we sure this won't add multiple probes to a Quarkus project?

rohanKanojia:
> I'd expect this Enricher also affecting Quarkus, but I didn't see any changes in that module.
Are we sure this won't add multiple probes to a Quarkus project?

Right now enrichers are executed in the order specified in profiles-default.yml:
```
      - jkube-healthcheck-quarkus
      - jkube-healthcheck-spring-boot
      - jkube-healthcheck-thorntail-v2
      - jkube-healthcheck-wildfly-jar
      - jkube-healthcheck-karaf
      - jkube-healthcheck-vertx
      - jkube-healthcheck-docker
      - jkube-healthcheck-webapp
      - jkube-healthcheck-micronaut
      - jkube-healthcheck-openliberty
      - jkube-healthcheck-smallrye
```

I've placed SmallRyeHealthCheckEnricher at the bottom. Once an enricher adds liveness/readiness probes other health check enrichers would simply ignore it. The logic is to add probes only if they don't exist.
https://github.com/eclipse/jkube/blob/9490349b58b730c9be6fea7c4701f1608ccccdc7/jkube-kit/enricher/specific/src/main/java/org/eclipse/jkube/kit/enricher/specific/AbstractHealthCheckEnricher.java#L63

manusa:
> I've placed SmallRyeHealthCheckEnricher at the bottom. Once an enricher adds liveness/readiness probes other health check enrichers would simply ignore it. The logic is to add probes only if they don't exist.

This is very brittle (especially since there is no test right now to verify the possible permutations and their effects, or is there?)

jkube-kit/jkube-kit-microprofile/src/main/java/org/eclipse/jkube/microprofile/MicroprofileHealthUtil.java
rohanKanojia: I've checked again with my test projet. It seems to be working okay.

rohanKanojia:
I don't think there is any test to enforce enricher ordering.

manusa:
> I don't think there is any test to enforce enricher ordering.

There isn't , but still relying just on the ordering of the config file is really dangerous.

I'm not sure what the collisions between the new enricher and previous ones are, but we should make sure that only one of the applicable is applied via some sort of scoring system.

rohanKanojia:
I think we should tackle this in a separate issue.

manusa:
> I think we should tackle this in a separate issue.

:+1: Please create an issue with a summary and link to this PR+issue

rohanKanojia:
I've created https://github.com/eclipse/jkube/issues/1890

PR #1525, opened 2022-05-27
fix : Wrong information in the docs about run-sh and other `io.fabric8` artifacts (#1006)

rohanKanojia:
## Description

Fix #1006

+ Fix outdated dmp plugin descriptor for build-plugins functionality. Introduce an interface `org.eclipse.jkube.kit.api.JKubeBuildPlugin` which would implemented by build plugins
+ rather than relying on on reflection, add an interface
    `org.eclipse.jkube.kit.common.JKubeBuildPlugin` which would be
    implemented by each build plugin. Instead of loading manually, use
    `java.util.ServiceLoader` to load the implementations automatically.
+ Quickstart added in https://github.com/eclipse/jkube/pull/1711
+ Docs added in https://github.com/eclipse/jkube/pull/1712

Signed-off-by: Rohan Kumar <rohaan@redhat.com>

## Type of change

 - [X] Bug fix (non-breaking change which fixes an issue)
 - [ ] Feature (non-breaking change which adds functionality)
 - [ ] Breaking change (fix or feature that would cause existing functionality to change
 - [ ] Chore (non-breaking change which doesn't affect codebase;
   test, version modification, documentation, etc.)

## Checklist
 - [X] I have read the [contributing guidelines](https://www.eclipse.org/jkube/contributing)
 - [X] I signed-off my commit with a user that has signed the [Eclipse Contributor Agreement](https://www.eclipse.org/legal/ECA.php)
 - [X] I Added [CHANGELOG](../CHANGELOG.md) entry
 - [X] I have implemented unit tests to cover my changes
 - [X] I have updated the [documentation](../kubernetes-maven-plugin/doc) accordingly
 - [ ] No new bugs, code smells, etc. in [SonarCloud](https://sonarcloud.io/dashboard?id=jkubeio_jkube) report
 - [X] I tested my code in Kubernetes
 - [ ] I tested my code in OpenShift

Issue #1006, opened 2021-10-13
Wrong information in the docs about run-sh and other io.fabric8 artifacts


grgrzybek:
### Info

I was looking at:
* https://www.eclipse.org/jkube/docs/kubernetes-maven-plugin
* https://www.eclipse.org/jkube/docs/openshift-maven-plugin

and I believe these are the transitioned versions of https://dmp.fabric8.io/#maven-goals

For example in the fabric8's docker-maven-plugin there's:
```
Build Plugins

This plugin supports so call dmp-plugins which are used during the build phase. dmp-plugins are enabled by just declaring a dependency in the plugin declaration:

<plugin>
  <groupId>io.fabric8</groupId>
  <artifactId>docker-maven-plugin</artifactId>

  <dependencies>
    <dependency>
      <groupId>io.fabric8</groupId>
      <artifactId>run-java-sh</artifactId>
      <version>1.2.2</version>
    </dependency>
  </dependencies>
</plugin>

These plugins contain a descriptor META-INF/maven/io.fabric8/dmp-plugin with class names, line-by-line:

io.fabric8.runsh.RunShLoader
```

While in jkube docs there's simply:
```
Build Plugins

This plugin supports so call dmp-plugins which are used during the build phase. dmp-plugins are enabled by just declaring a dependency in the plugin declaration:

<plugin>
  <groupId>org.eclipse.jkube</groupId>
  <artifactId>kubernetes-maven-plugin</artifactId>

  <dependencies>
    <dependency>
      <groupId>org.eclipse.jkube</groupId>
      <artifactId>run-java-sh</artifactId>
      <version>1.2.2</version>
    </dependency>
  </dependencies>
</plugin>

org.eclipse.jkube.runsh.RunShLoader
```

while `kubernetes-maven-plugin` artifact _was_ transitioned from `io.fabric8` to `org.eclipse.jkube` groupId, it isn't the case with `io.fabric8/run-java-sh` GA and `io.fabric8.runsh.RunShLoader` class...

rohanKanojia:
We do have functionality ported from dmp for this. However, DMP_PLUGIN_DESCRIPTOR constant is wrongly configured due to which this functionality doesn't work at the moment. 

https://github.com/eclipse/jkube/blob/40added43dc6fbb180e2c79572628d0471be4195/kubernetes-maven-plugin/plugin/src/main/java/org/eclipse/jkube/maven/plugin/mojo/build/AbstractDockerMojo.java#L101

[...]

PR #1711, opened 2022-08-19
feat (quickstarts/kit) : Add Kit Quickstart for implementing a custom JKube Plugin (#1006)

rohanKanojia:
## Description

- [X]   Requires #1525 to be merged first.

Add Custom Plugin quickstart for implementing custom JKube plugin.

## Type of change

## Checklist


manusa:
IMHO the quickstart should belong to the Maven section, since it's a multi-module Maven project showing how to create the plugin, but also how to use it in Maven.

Consider how we need to add a specific suffix to be able to provide quickstarts both for Maven and for Gradle.

_Same should probably apply to the custom-istio-enricher examples (--> we should create a technical-debt issue)_

quickstarts/kit/custom-build-plugin/README.md
```diff
@@ -0,0 +1,10 @@
+---
+name: "Kit :: Maven :: Custom Build Plugin Parent"
+description: |
+A simple project demonstrating how to write your own custom build plugins using Eclipse JKube.
+One module is build plugin and other is app module consuming the custom build plugin.
+---
+
+This project contains two submodules:
+- [App](./app)
+- [Foo Build Plugin](./foo-build-plugin)
```
manusa, line 10: I think it might be reasonable to merge both READMEs in this one with appropriate headings (H2).

It will be easier to follow for users.

manusa: I'd also repurpose the build plugin and the consuming app.

We could use the build plugin to retrieve data from an online service and populate a file.
This file, or its contents, could then be served by the final application. So each time the application was built and deployed, the file would be regenerated. This is exactly what we're doing (using JavaScript) in the JKube website when we process the asciidoc documentation, or the quickstarts.

In general, this would resemble more what a real use-case would look like.

For example, a real use case that would be useful for us (please don't implement this) in JKube would be script that packaged a single quickstart in a zip file so that it could be then downloaded from the main application web service.

manusa (changes requested):
If you want we can further discuss the purpose of the Foo plugin, but I think that creating something cool and demo it would be good to drive attention to JKube.

If you want me to implement a draft of something else, let me know.

manusa (approved):

Issue #1783, opened 2022-10-05
[Hacktoberfest] AuthConfigFactory fields may be 'final'


manusa:
### Description
The following statement is reported as a code style issue:
 
https://github.com/eclipse/jkube/blob/17ada413b53017ce446da3326ef1a5ff31d48e43/jkube-kit/build/service/docker/src/main/java/org/eclipse/jkube/kit/build/service/docker/auth/AuthConfigFactory.java#L638
 
- [x] should be changed to:
```java
        private final String configMapKey;
```
 
### Before you start
 
Make sure you read the [contributing guide](https://www.eclipse.org/jkube/contributing) first. Pay special attention to the ECA agreement section and the requirement to sign-off your commit.
 
#### Kubernetes
 
If you don't have a real Kubernetes cluster available (most probably), you can use Minikube or Kind to test with a local cluster.
 
- Minikube [get started guide](https://minikube.sigs.k8s.io/docs/start/)
- Kind (Requires Docker) [quick start](https://kind.sigs.k8s.io/docs/user/quick-start/)
 
#### OpenShift
 
If you don't have a real OpenShift cluster available (most probably), you can use [Red Hat's developer Sandbox for Red Hat OpenShift](https://developers.redhat.com/developer-sandbox). The only requirement is to have a Red Hat account.
 
Once you have your Sandbox environment, you'll need to download the oc tool from the cluster console.
(Press the `?` icon and from the context menu select `Command line tools`, you'll be redirected to https://$subdomain.openshiftapps.com/command-lines-tools where you'll be able to download the CLI for your platform)

shubhamrawat090:
hey @manusa is anyone working on this? please assign it to me if not

ishu-thakur:
Hi @manusa can you assign that to me :) ?

DevPJ9:
Hello @manusa I would like to work on this one please assign this to me.

Gairick52:
Hello @manusa i would love to work on this one please assign this to me

manusa:
> Hello @manusa i would love to work on this one please assign this to me

Hi @Gairick52 
This one is already taken, but you can find similar issues labeled with [hacktoberfest](https://github.com/eclipse/jkube/labels/Hacktoberfest) still available.

DevPJ9:
Hey, @manusa I have raised a PR. Can you please review it?

Pawank06:
Hey, @manusa can you please assign me this issue?

[...]

PR #1790, opened 2022-10-05
AuthConfigFactory fields may be 'final' #1783

ishu-thakur:
Changing the Scope of Fields in the AuthConfigFactory  issue #1783


manusa:
Hi @ishu-thakur 
You've already contributed to the project, issues labeled with `first-timers-only` are limited to first time contributors to the project (and at most one).
The whole point of these very simple issues is for people to get acquainted with the project, with the Eclipse Foundation processes, and in general with free open source software. Consider that it takes longer for me to create the issue than to create the actual fix. We believe that these kind of issues help the community and this is why we put an extra-effort submitting them, if a single contributor was to tackle all of these issues, our effort would be wasted.

Issue #1781, opened 2022-10-05
[Hacktoberfest] ArchiveService.assemblyManager may be 'final'


manusa:
### Description
The following statement is reported as a code style issue:
 
https://github.com/eclipse/jkube/blob/4b7998d19509b1fbc73865a2a19a0112c3a21e1c/jkube-kit/build/service/docker/src/main/java/org/eclipse/jkube/kit/build/service/docker/ArchiveService.java#L35
 
- [ ] should be changed to:
```java
    private final AssemblyManager assemblyManager;
```
 
### Before you start
 
Make sure you read the [contributing guide](https://www.eclipse.org/jkube/contributing) first. Pay special attention to the ECA agreement section and the requirement to sign-off your commit.
 
#### Kubernetes
 
If you don't have a real Kubernetes cluster available (most probably), you can use Minikube or Kind to test with a local cluster.
 
- Minikube [get started guide](https://minikube.sigs.k8s.io/docs/start/)
- Kind (Requires Docker) [quick start](https://kind.sigs.k8s.io/docs/user/quick-start/)
 
#### OpenShift
 
If you don't have a real OpenShift cluster available (most probably), you can use [Red Hat's developer Sandbox for Red Hat OpenShift](https://developers.redhat.com/developer-sandbox). The only requirement is to have a Red Hat account.
 
Once you have your Sandbox environment, you'll need to download the oc tool from the cluster console.
(Press the `?` icon and from the context menu select `Command line tools`, you'll be redirected to https://$subdomain.openshiftapps.com/command-lines-tools where you'll be able to download the CLI for your platform)

ankitjc:
I can take this up.

PR #1793, opened 2022-10-05
Changing the Scope of fields in the ArchiveService to final issue #1781.

ishu-thakur:
Changing the Scope of fields in the ArchiveService to final issue #1781.

### DIFF
--- a/jkube-kit/build/service/docker/src/main/java/org/eclipse/jkube/kit/build/service/docker/ArchiveService.java
+++ b/jkube-kit/build/service/docker/src/main/java/org/eclipse/jkube/kit/build/service/docker/ArchiveService.java
@@ -32,7 +32,7 @@ import java.util.List;
 public class ArchiveService {
 
     private final KitLogger log;
-    private AssemblyManager assemblyManager;
+    private final AssemblyManager assemblyManager;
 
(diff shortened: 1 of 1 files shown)


manusa:
Hi @ishu-thakur 
You've already contributed to the project, issues labeled with `first-timers-only` are limited to first time contributors to the project (and at most one).
The whole point of these very simple issues is for people to get acquainted with the project, with the Eclipse Foundation processes, and in general with free open source software. Consider that it takes longer for me to create the issue than to create the actual fix. We believe that these kind of issues help the community and this is why we put an extra-effort submitting them, if a single contributor was to tackle all of these issues, our effort would be wasted.

ishu-thakur:
Hi @manusa yes sure i understand that :) so if there is any issue that is not first time only but beginner as i am new to contribution i would like to contribute , and sorry i just didnt know about it thats why i raised the PR by fixing the issues but i will keep this time in my mind and thanks for guiding me :).

manusa:
No worries, we usually tag these issues with [`help wanted`](https://github.com/eclipse/jkube/labels/help%20wanted). There are not many right now our time for these tasks is absorbed by GSoC and Hacktoberfest.

PR #1888, opened 2022-11-04
feat(gradle): KubernetesExtension has helper method to add image with builder

manusa:
## Description

Adds a convenience method to the `KubernetesExtension` to allow adding images using a builder:

```groovy
kubernetes {
  addImage {builder ->
    builder.name = 'the name'
  }
}
```

I was wondering if we would prefer something like the following snippet instead:
```groovy
kubernetes {
  images {
    imageBuilder()
      .name('the-name')
      .build(imageBuildBuilder()./*...*/build())
  }
}
```

/cc @melix

## Type of change

## Checklist
 - [x] I have added tests that prove my fix is effective or that my feature works
 - [x] New and existing unit tests pass locally with my changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ Usage:
 ```
 ### 1.10-SNAPSHOT
 * Fix #1684: Podman builds with errors are correctly processed and reported 
+* Fix #1888: KubernetesExtension has helper method to add image with builder
 
 ### 1.9.1 (2022-09-14)
 * Fix #1747: Apply service doesn't attempt to create OpenShift Projects in Kubernetes clusters
--- a/gradle-plugin/it/src/it/groovy-dsl-image/build.gradle
+++ b/gradle-plugin/it/src/it/groovy-dsl-image/build.gradle
@@ -1,3 +1,4 @@
+
 plugins {
     id 'org.eclipse.jkube.kubernetes' version "${jKubeVersion}"
     id 'org.eclipse.jkube.openshift' version "${jKubeVersion}"
@@ -12,6 +13,7 @@ repositories {
     mavenCentral()
 }
 
+import org.eclipse.jkube.kit.config.image.build.BuildConfiguration
 kubernetes {
     offline = true
     images {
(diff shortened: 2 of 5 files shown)


melix:
Thanks @manusa ! It will be of great interest for both Kotlin DSL users and people integrating your plugin into theirs (in plain Java).

> I was wondering if we would prefer something like the following snippet instead:

I don't think you should. It's not Gradle idiomatic. Note that the current way of configuring is not Gradle idiomatic either. In Gradle, you should have properties which are mutable for the whole configuration phase. This is done so that plugins can provide conventional values, which can be overwritten by user configuration in build scripts, or other plugins (written in Java, Groovy, ...). Currently, your plugin is using the structures from the toolkit API which are immutable. This means that you assume that one the image is created (or any other nested element), its configuration is _done_ and cannot be changed. This is not something that you should assume with Gradle. This makes this plugin particularly difficult to work with.

By using a builder like in the `addImage` method, we're _closer_ to what Gradle configuration should look like (mutable data in build scripts), but obviously not like what I described: in the end, you'll store an _immutable_ image. This means that in practice we will be able to provide a _default_ image for users, but they wouldn't be able to tweak its parameters for their own needs. This _may_ be enough for us, but I don't know at this stage.

Eventually, I would recommend that you have 2 separate data structures: one mutable, for configuration time, and one _immutable_, that you use at execution time. I reckon that this may be significant rework. I actually tried to give it a shot, but figured out that it would be difficult because of the use of Lombok.

In any case, I appreciate that you're considering adding such a method, which will at least unblock our experiments.

manusa:
> Thanks @manusa ! It will be of great interest for both Kotlin DSL users and people integrating your plugin into theirs (in plain Java). [...]

Thanks for your detailed explanation :raised_hands: 

> Eventually, I would recommend that you have 2 separate data structures: 

When we started the Gradle plugins implementation, we didn't want to pollute the jkube-kit modules with Gradle-specific annotations and types. That's why we're mostly using the closure-deserialization hack to initialize the objects.
We also wanted to keep the configuration classes at a single point to simplify their maintenance. We could consider the usage of a DTO-like process to provide Gradle specific configuration classes that could later on be converted/mapped into the main jkube-kit configuration types.

Another thing that I'm not sure if you've tried, but could maybe be a middle-ground is using the `.toBuilder()` method that most of our configuration classes provide. Although tedious, this should provide a possibility to override any of the provided defaults.


Anyway, this PR should be released in 1.10.0 which should be out by next week :crossed_fingers:

rohanKanojia (approved):

Issue #523, opened 2020-12-14
dot in repository delete the registry configuration in pom.xml


rgordill:
### Description
When using dot in repositories, the registry part is missing. Example:
```
<name>rgordill/%a:%l</name>
<registry>quay.io</registry>
```
Gives a correct images in the deployment: 
`image: quay.io/rgordill/customer-service-cache:latest`

But when using:
```
<name>ramon.gordillo/%a:%l</name>
<registry>quay.io</registry>
```
Gives a correct images in the deployment: 
`image: ramon.gordillo/customer-service-cache:latest`

### Info

* Eclipse JKube version : 1.0.2
* Maven version (`mvn -v`) : 3.6.3

* Kubernetes / Red Hat OpenShift setup and version : k8s 1.19

* If it's a bug, how to reproduce : See the reproducer project

* Sample Reproducer Project : *https://github.com/rgordill/customer-service-cache/tree/af19d20254895892ce184a672cd35364b13d4ddf*

manusa:
Relates to: https://github.com/eclipse/jkube/issues/526

rohanKanojia:
Quay.io Seems to allow a username with dots:
![Screenshot_20210121_192446](https://user-images.githubusercontent.com/13834498/105360812-02769700-5c1f-11eb-94ae-ed2ab414eef9.png)

However, DockerHub doesn't allow such usernames:
![Screenshot_20210121_192714](https://user-images.githubusercontent.com/13834498/105360853-11f5e000-5c1f-11eb-87b4-714aa19c7026.png)

I think this is due to an expectation in `ImageName` class that is there is a `.` or `:` in name then it's a registry:
https://github.com/eclipse/jkube/blob/ff0bbc82a1b2a9a05465a6e62a35f4184a2c7c3d/jkube-kit/config/image/src/main/java/org/eclipse/jkube/kit/config/image/ImageName.java#L145-L148

Due to this assumption, we're assuming the first one is a registry and assign it to the registry in image name here:
https://github.com/eclipse/jkube/blob/ff0bbc82a1b2a9a05465a6e62a35f4184a2c7c3d/jkube-kit/config/image/src/main/java/org/eclipse/jkube/kit/config/image/ImageName.java#L278-L283

It gets called via `ContainerHandler` during resource phase here:
[...]

PR #2011, opened 2023-01-21
#1929 Fixed parsing image name

### LINKED ISSUE #1929: Pull registry is not getting authenticated
### Describe the bug

`<sever>` authentication in `settings.xml` doesn't work for pullregistry.
Also using docker login is not respected with pullregistry (settings.xml without server section).

I have two separate docker registries. One for pull (dockerhub proxy) and one for push.
Case 1. jkube doesn't respect `settings.xml` configuration 

settings.xml
```
<settings>
  <servers>
    <server>
      <id>my.proxy</id>
      <username>user</username>
      <password>pass</password>
    </server>
  </servers>
</settings>
```
Dockerfile
```
FROM my.proxy/openjdk:17
# rest of the file
```
I noticed that everything works as expected when I set server id to docker.io 
Case 2. After logging via docker login to my.proxy I'm getting unauthorized exception.

### Eclipse JKube version

1.10.0

### Component

Kubernetes Maven Plugin

### Apache Maven version

3.8.1

### Gradle version

_No response_

### Steps to reproduce

1. Set settings.xml
```
<settings>
  <servers>
    <server>
      <id>my.proxy</id>
      <username>user</username>
      <password>pass</password>
    </server>
  </servers>
</settings>
```
2. Set docker file
Dockerfile
```
FROM my.proxy/openjdk:17
```
3. Set pull registry in pom:
```
 <plugin>
        <version>1.10.1</version>
        <configuration>
          <pullregistry>my.proxy</pullregistry>
        </configuration>
```

### Expected behavior

Image is being downloaded from registry.

### Runtime

Kubernetes (vanilla)

other (please specify in additional context)

### Environment

macOS

### Eclipse JKube Logs

```shell
Error while trying to build the image: Unable to pull 'my.proxy/openjdk:17' : {"message":"Head \"https://my.proxy/v2/openjdk/manifests/17\": unknown: Authentication is required"} (Internal Server Error: 500) -> [Help 1]
```

### Sample Reproducer Project

_No response_

### Additional context

_No response_

balbusm:
## Description
Fixes #1929 
Image name (foo-bar-registry.jfrog.io/java:jre-17) was incorrectly parsed what caused failed on image pull.
foo-bar-registry.jfrog.io was parsed as user instead of registry.
Changed the way docker image name is parsed:
- User part for docker hub can contain only alphanumeric values
- Image - foo-bar-registry.jfrog.io/java:jre-17 - parses foo-bar-registry.jfrog.io as repository
- Image - eclipse/eclipse_jkube parses eclipse as user, hence image is treated as docker hub image

## Type of change

## Checklist
 - [x] I have added tests that prove my fix is effective or that my feature works
 - [x] New and existing unit tests pass locally with my changes


jkube-kit/enricher/api/src/test/java/org/eclipse/jkube/kit/enricher/handler/ContainerHandlerTest.java
```diff
@@ -195,13 +195,13 @@ void withImageNames_shouldReturnContainersWithConfiguredImages() {
       }
 
       @Test
-      @DisplayName("with user, image and tag with period in image user, should return configured image")
-      void withUserAndImageAndTagWithPeriodInImageUser_shouldReturnConfiguredImage() {
+      @DisplayName("with user, image and tag with dash in image user, should return configured image")
+      void withUserAndImageAndTagWithDashInImageUser_shouldReturnConfiguredImage() {
```
rohanKanojia, line 199: Is it necessary to modify existing tests? I remember this test was related to https://github.com/eclipse/jkube/issues/523 in which the user name contained `.` 

You can add additional test to cover your use case.

sunix: I think it is not correct to have a . in the "username" part

balbusm: I think problem is more complicated.
Basically I see those cases:
- ecplise_jkube - there is one part, it's a docker hub registry
- eclipse/ecplise_jkube - there are 2 parts; it's docker hub registry; eclipse repo
- user.jfrog/eclipse_jkube - there are 2 parts; it's jfrog registry;
- quay.io/my.user/eclipse_jkube - there are 3 parts; it's quay.io registry; my.user username
I think my fix is correct.
Username can have dots if it's not a docker hub registry. Based on dot in 2 parts image name it's possible to determine if it's registry or username.
3 parts image name can have dots.
However there is a logic that adds registry after image name was parsed (eg ImageName.getFullName(String optionalRegistry)). We cannot add registry once image name was parsed as it's not possible to determine if we dealing with registry or username.
ImageName.getFullName is used in combination with EnvUtil.firstRegistryOf(imageName.getRegistry(), registryConfig.getRegistry());
How can it parse ImageName properly if registry is selected based on ImageName or RegistryConfig?

BTW there is another case
- user.jfrog/my-repo/eclipse_jkube - there are 3 parts; it's jforg registry; user is a username; my-repo is a repo
I would park this case as user can explicitly provide username

sunix (approved):
It looks good to me,
however I feel like this part is a bit fragile, I would refactor this with a regexp similar to https://github.com/eclipse-che/che-plugin-registry/blob/main/build/dockerfiles/entrypoint.sh#L39, but maybe this could be part of another PR.

jkube-kit/enricher/api/src/test/java/org/eclipse/jkube/kit/enricher/handler/ContainerHandlerTest.java
manusa: > I think it is not correct to have a . in the "username" part

We did have an issue (#523) and needed to provide a fix for this specific case `roman.fake/customer-service-cache:latest` 

So we need to ensure that the fix provided for the scenario described in the issue is still applicable

sunix: https://docs.docker.com/engine/reference/commandline/tag/

> An image name is made up of slash-separated name components, optionally prefixed by a registry hostname. The hostname must comply with standard DNS rules, but may not contain underscores. If a hostname is present, it may optionally be followed by a port number in the format :8080. If not present, the command uses Docker’s public registry located at registry-1.docker.io by default. Name components may contain lowercase letters, digits and separators. A separator is defined as a period, one or two underscores, or one or more dashes. A name component may not start or end with a separator.
> 
> A tag name must be valid ASCII and may contain lowercase and uppercase letters, digits, underscores, periods and dashes. A tag name may not start with a period or a dash and may contain a maximum of 128 characters.

balbusm: Then there is no way to determine if it's a username or service domain name.
repo.jfrog.io/eclipse_jkube vs roman.fake/eclipse_jkube 
I see below solutions:
- add special case for jfrog parsing
- do some extra logic to do some fall back logic for download
- add section to the documentation and error message to explicitly add port number eg repo.jfrog.io:80/eclipse_jkube

rohanKanojia: @manusa @sunix : thoughts??

sunix: in Dockerhub from what I see (when creating an account), a username can have `4 to 30 letters & digits only`.

sunix: In all the cases, I am wondering how clients like `docker` or `podman` behave in each cases. We should rely on it maybe.

rohanKanojia:
@balbusm : I would like to move forward with this fix. But we also don't want to break existing tests. Shall I take over your PR to find some middle ground so that we don't break existing tests and cover your use case as well?

PR #2183, opened 2023-05-30
fix (jkube-kit/config/image) : Remove user field from ImageName (#2116)

### LINKED ISSUE #2116: docker basic Auth failure
### Describe the bug

I am trying to build and push my image. I have a local docker registry configured and working fine with docker cmdline interactions.
For auth I have gone with a simple server section in the ~/.m2/settings.yml  
```
    <servers>               
        <server>
            <id>myregistry.127.0.0.1.nip.io</id>
            <username>user</username>
            <password>pwd123</password>
        </server>             
    </servers>
```
the base image is called myregistry.127.0.0.1.nip.io/eclipse-temurin:11.0.18_10-jdk-alpine
When I run k8s:build for example the build fails pulling the base image "no basic auth".
If I pass the docker creds in on the cmdline it works so its like its not matching the registry in the image to the registry in the server section in the settings.xml

any advice would be appreciated

### Eclipse JKube version

1.11.0

### Component

Kubernetes Maven Plugin

### Apache Maven version

3.8.5

### Gradle version

None

### Steps to reproduce

as above

### Expected behavior

image should be pulled from registry

### Runtime

Kubernetes (vanilla)

1.25.3

### Environment

Linux

### Eclipse JKube Logs

_No response_

### Sample Reproducer Project

_No response_

### Additional context

_No response_

rohanKanojia:
Fix #2116
Closes #2011

Relates to:
- #523 
- #543
- #2011

Earlier we split image names before tagging into `registry`, `user` and `repository`. In Docker code, there is no mention of this user attribute. Split image name into `registry` and `repository` only. User can be extracted from `repository` if required. This would mean that in some cases(`foo.bar/name:tag`) path element would be considered as registry, added a case in getNameWithoutTag to consider optionalRegistry as registry.

+ Remove user field from ImageName class
   - In Docker reference package, there is no mention of user field. A valid
repository is split into a domain and path component. Remove this user
field, but keep getUser method which will try to extract possible
username from path component (repository).
    - Add `isFullyQualifiedName` method to ImageName so that we can distinguish between image names of complete notation ($registry/$user/$name:$tag) from incomplete ones
    - There are some places in code where registry gets appended to image name later using `getFullName(optionalRegistry)`. If given image name is not fully qualified and registry element is a valid path element, then `getNameWithoutTag` would consider `optionalRegistry` as registry and current registry as path element

## Type of change

## Checklist

### DIFF
--- a/jkube-kit/config/image/src/main/java/org/eclipse/jkube/kit/config/image/ImageName.java
+++ b/jkube-kit/config/image/src/main/java/org/eclipse/jkube/kit/config/image/ImageName.java
@@ -142,16 +142,8 @@ public class ImageName {
         return builder.toString();
     }
 
-    private boolean containsPeriodOrColon(String part) {
-        return containsPeriod(part) || containsColon(part);
-    }
-
-    private boolean containsPeriod(String part) {
+    private boolean isRegistryPart(String part) {
         return part.contains(".");
-    }
-
-    private boolean containsColon(String part) {
-        return part.contains(":");
     }
 
     /**
@@ -281,26 +273,18 @@ public class ImageName {
             registry = null;
             user = null;
             repository = parts[0];
-        } else if (parts.length >= 2) {
-            if (containsPeriodOrColon(parts[0])) {
-                if (parts.length > 2) {
-                    assignRegistryUserAndRepository(parts);
-                } else {
-                    checkWhetherFirstElementIsUserOrRegistryAndAssign(parts);
-                }
+        } else if (parts.length == 2) {
+            if (isRegistryPart(parts[0])) {
+                assignRegistryAndRepository(parts);
             } else {
-                registry = null;
-                user = parts[0];
-                repository = rest;
+                assignUserAndRepository(parts);
             }
-        }
-    }
-
-    private void checkWhetherFirstElementIsUserOrRegistryAndAssign(String[] parts) {
-        if (containsColon(parts[0])) {
-            assignRegistryAndRepository(parts);
-        } else {
-            assignUserAndRepository(parts);
+        } else if (parts.length >= 3) {
+            if (isRegistryPart(parts[0])) {
+                assignRegistryUserAndRepository(parts);
+            } else {
+                assignUserAndRepository(parts);
+            }
         }
     }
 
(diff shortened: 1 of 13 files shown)


jkube-kit/config/image/src/main/java/org/eclipse/jkube/kit/config/image/ImageName.java
```diff
@@ -220,7 +232,13 @@ public String getFullName(String optionalRegistry) {
      * @return user part or <code>null</code> if no user is present in the name
      */
     public String getUser() {
```
manusa, line 234: Not following completely. Is `user` as an entity relevant for the image name?
The PR says `Remove user field from ImageName`, however we just seem to  be delaying its computation

rohanKanojia: I don't think it's relevant for image name. There are a few places in codebase where we seem to be checking for image user:

- AuthConfigFactory: while fetching credentials from settings.xml, we check for server attributes whether they have same id as registry name or ${registry}/${user}
- KubernetesResourceUtil: while extracting container name, if alias is not provided then either ${user}-${project.artifactId} or ${project.groupId}-${project.artifactId} is used
TriggersAnnotationEnricher
- BaseGenerator: while setting JKubeBuildStrategy.SourceStrategy.namespace attribute
- TriggersAnnotationEnricher: to decide whether add triggers annotation or not

I just kept this method for these places to work.

manusa: OK, would need to check those methods. But is it relevant to call this `user`? Meaning, is this acting as a user (name) in any of the mentioned cases? If not, maybe we can rename it to call it something more relevant, else it will continue to cause confusion

manusa: No it still seems to be unrelated to the image name.
I'll try to check out exactly how the method is consumed and see if I can propose any alternatives.

manusa:
I added a further commit.

I finally renamed the getUser field to inferUser to avoid the confusion where we might be led think that a `user` is actually something meaningful from the Docker's image name naming perspective.
In the future, we should probably revisit where we're actually using this and maybe move that logic somewhere else, where it is more relevant to its purpose.

I also separated the tests and provided clear, explicit, cases where we can actually understand the results of how the image name is decomposed into several parts (we can also see the process in the test case).

manusa (approved):

manusa:
Will this one close #2011 ?

rohanKanojia:
yes, I had cherry-picked commit from that PR onto this one.

Issue #2400, opened 2023-09-28
Helm: support for complex values


manusa:
## Description

Follow up on #2200
Requires #2402 

Be able to provide through Helm values fragments complex values such as annotations or arrays.

```yaml
# values.helm.yaml
annotations:
  example.org/one: one
  example.org/two: two
```

With a fragment such that
```yaml
# deployment.yaml
metadata:
  annotations:
    ${annotations}
```

Can be processed by providing a Helm parameter:

```groovy
// build.gradle
kubernetes {
  helm {
    parameters = [{
      name = 'annotations'
      value = '{{- toYaml .Values.annotations | nindent 4 }}'
    }]
  }
}
```

PR #2404, opened 2023-09-28
fix: helm no longer generates default function; broadens support for different value types

### LINKED ISSUE #2399: Helm: No need for the `default` function
## Description

Currently, we're computing a Helm expression for provided parameters where a [`default`](https://helm.sh/docs/chart_template_guide/functions_and_pipelines/#using-the-default-function) function is added in case the value is provided.

In our case, this default function is redundant since the value will also be added to the generated `values.yaml` file.

The `default` function is problematic when dealing with non-string variables.

Since the function is redundant, it's better to just remove it from the generated helm expression.

manusa:
## Description

Fixes #2399

fix: helm no longer generates default function; broadens support for different value types

## Type of change

 - [x] Breaking change (fix or feature that would cause existing functionality to change

## Checklist
 - [x] I have read the [contributing guidelines](https://www.eclipse.dev/jkube/contributing)

### DIFF
--- a/jkube-kit/resource/helm/src/test/resources/it/expected/openshift/values.yaml
+++ b/jkube-kit/resource/helm/src/test/resources/it/expected/openshift/values.yaml
@@ -13,9 +13,9 @@
 #
 
 deployment:
-  replicas: "1"
+  replicas: 1
   limits:
-    memory: 512Mi
+    memory: "512Mi"
 GLOBAL_TEMPLATE_ENV_VAR: This is a sample
 unused_value: the-value
 requests_memory: 256Mi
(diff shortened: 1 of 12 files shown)


jkube-kit/resource/helm/src/test/resources/it/expected/openshift/values.yaml
```diff
@@ -13,9 +13,9 @@
 #
 
 deployment:
-  replicas: "1"
+  replicas: 1
   limits:
-    memory: 512Mi
+    memory: "512Mi"
```
rohanKanojia, line 18: Why have you wrapped it in quotes in `expected/openshift/values.yaml` but not in `expected/kubernetes/values.yaml`?

rohanKanojia (approved):

jkube-kit/resource/helm/src/test/resources/it/expected/openshift/values.yaml
manusa: For this value it doesn't really make sense because logic says that the value should be deserialized as a String. The goal here is to clearly highlight that this is treated as a String, while other values (such as replicas) can be treated as integers.

This was not the case before, when everything was dealt as a String.
This will become more clear with follow up PRs regarding #2400

PR #2405, opened 2023-09-28
fix (jkube-kit) : lastModified timestamps should be preserved while preparing assembly

rohanKanojia:
## Description

We don't seem to be preserving lastModified timestamps of files that are being copied while preparing the assembly for image. Make sure we're copying all the metadata of files/directories as well while copying them.

## Type of change

## Checklist

### DIFF
--- a/jkube-kit/common/src/main/java/org/eclipse/jkube/kit/common/util/FileUtil.java
+++ b/jkube-kit/common/src/main/java/org/eclipse/jkube/kit/common/util/FileUtil.java
@@ -36,6 +36,7 @@ import java.util.Objects;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import static java.nio.file.StandardCopyOption.COPY_ATTRIBUTES;
 import static java.nio.file.StandardCopyOption.REPLACE_EXISTING;
 
 /**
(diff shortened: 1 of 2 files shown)


jkube-kit/common/src/main/java/org/eclipse/jkube/kit/common/archive/AssemblyFileSetUtils.java
```diff
@@ -169,8 +169,10 @@ private static List<AssemblyFileEntry> copy(Path sourceDirectory, File source, F
         for (File sourceChild : Optional.ofNullable(source.listFiles()).orElse(new File[0])) {
           copy(sourceDirectory, sourceChild, new File(target, sourceChild.getName()), assemblyFileSet);
         }
+        FileUtil.updateLastModifiedTimestamp(target, source.lastModified());
```
manusa, line 172: I think it'd be better if this was done as part of the copy.
Instead of adding an extra-function, just modify FileUtil.copy to preserve the original metadata

manusa: OK, in this case target refers to the directoy.
Likewise, whatever process was used to create the directory should be used to preserve that timestamp.
I don't see the logic (besides for this current use-case) in updating the timestamp of an existent directory.

rohanKanojia: Umm, you mean FileUtil.copy?

I noticed that whenever a new file gets added to directory, it's lastModified timestamp gets updated. This is done to revert directory's lastModified to original source's lastModified

manusa (approved):

PR #2420, opened 2023-10-11
fix (jkube-kit) : Fix `Fabric8HttpUtil.extractAuthenticationChallengeIntoMap` www-authenticate header parsing logic (#2419)

### LINKED ISSUE #2419: Getting `Unauthorized` while pushing helm chart to OCI registry if repository name contains `push`
### Describe the bug

For a valid registry requiring authentication, I kept getting `Unauthorized` error even if I provided the correct credentials. When I debugged the issue, I discovered it had to do with the repository name. My test project name was `jkube-helm-push-oci` which affected this statement:

https://github.com/eclipse/jkube/blob/cc6d239259aab437655a8bfdb76f0328f1170a95/jkube-kit/resource/helm/src/main/java/org/eclipse/jkube/kit/resource/helm/oci/OCIRegistryInterceptor.java#L98-L99

While requesting token we need to add scope of that token : whether it's for a read (`pull`) or write operation (`push`). By default in first request Www-Authenticate header contains `pull` for read only. We need to manually add `push` to this in order to fetch token for uploading chart. However, this condition turns `false` if the repository name already contains `push` substring.

Rather than comparing scope field in Www-Authenticate header as plain string, we should parse it and then compare only the last segment.

### Eclipse JKube version

SNAPSHOT

### Component

Kubernetes Maven Plugin

### Apache Maven version

None

### Gradle version

None

### Steps to reproduce

Try pushing helm chart to some OCI registry with a chart name containing `push` keyword. Push would always fail

### Expected behavior

Push should be successful when valid credentials are provided regardless of project name

### Runtime

Kubernetes (vanilla)

1.25.3

### Environment

Linux

### Eclipse JKube Logs

```shell
[ERROR] k8s: Error performing helm push: java.lang.IllegalStateException: Failure in initiating upload request: Unauthorized
[INFO] BUILD FAILURE
[INFO] Total time:  6.903 s
[INFO] Finished at: 2023-10-11T12:02:04+05:30
[ERROR] Failed to execute goal org.eclipse.jkube:kubernetes-maven-plugin:1.15-SNAPSHOT:helm-push (default-cli) on project jkube-helm-push-testing: Failure in initiating upload request: Unauthorized -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoExecutionException
```

### Sample Reproducer Project

_No response_

### Additional context

_No response_

rohanKanojia:
## Description

Fixes #2419

We try to add push scope to scope field extracted from Www-Authenticate header if it doesn't already exist. This was happening because our `www-authenticate` header parsing logic was flawed and it was omitting `push` scope from `pull,push` scope. We were directly splitting whole string based on commas.

Port OkHttp's [HttpHeaders.parseChallengeHeader](https://github.com/square/okhttp/blob/parent-3.14.9/okhttp/src/main/java/okhttp3/internal/http/HttpHeaders.java#L180) method into our codebase for achieving this. We have runtime dependency on OkHttp, a transitive dependency coming from Fabric8 Kubernetes Client. We needed only `parseChallengeHeader` method for parsing WWW-Authenticate header for fetching authentication challenges. It seemed a better option to port the required method instead of adding okhttp dependency to the project.

## Type of change

## Checklist

### DIFF
--- a/jkube-kit/common/src/main/java/org/eclipse/jkube/kit/common/util/Fabric8HttpUtil.java
+++ b/jkube-kit/common/src/main/java/org/eclipse/jkube/kit/common/util/Fabric8HttpUtil.java
@@ -13,16 +13,16 @@
  */
 package org.eclipse.jkube.kit.common.util;
 
-import io.fabric8.kubernetes.client.http.HttpResponse;
-import org.apache.commons.lang3.StringUtils;
-
+import java.io.IOException;
 import java.io.UnsupportedEncodingException;
 import java.net.URLEncoder;
 import java.nio.charset.StandardCharsets;
-import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 
-import static org.apache.commons.lang3.StringUtils.strip;
+import io.fabric8.kubernetes.client.http.HttpResponse;
+
+import static org.eclipse.jkube.kit.common.util.okhttp.HttpHeaders.parseWwwAuthenticateChallengeHeaders;
 
 public class Fabric8HttpUtil {
   private static final String WWW_AUTHENTICATE = "WWW-Authenticate";
@@ -30,28 +30,14 @@ public class Fabric8HttpUtil {
   private Fabric8HttpUtil() { }
 
   /**
-   * Parse WWW-Authenticate Header as map
+   * Parse WWW-Authenticate Header challenges as map
    *
    * @param response Http Response of a particular request
-   * @return map containing various components of header as key value pairs
+   * @return list of maps containing various authentication challenges which each authentication challenge as map
    */
-  public static Map<String, String> extractAuthenticationChallengeIntoMap(HttpResponse<?> response) {
-    Map<String, String> result = new HashMap<>();
-    String wwwAuthenticateHeader = response.header(WWW_AUTHENTICATE);
-    if (StringUtils.isNotBlank(wwwAuthenticateHeader)) {
-      String[] wwwAuthenticateHeaders = wwwAuthenticateHeader.split(",");
-      for (String challenge : wwwAuthenticateHeaders) {
-        if (challenge.contains("=")) {
-          String[] challengeParts = challenge.split("=");
-          if (challengeParts.length == 2) {
-            result.put(challengeParts[0], strip(challengeParts[1], "\""));
-          }
-        }
-      }
-    }
-    return result;
+  public static List<Map<String, String>> extractAuthenticationChallengeIntoMap(HttpResponse<?> response) throws IOException {
+    return parseWwwAuthenticateChallengeHeaders(response.header(WWW_AUTHENTICATE));
   }
-
 
   /**
    * Create Form Data String from map
--- a/jkube-kit/common/src/test/java/org/eclipse/jkube/kit/common/util/Fabric8HttpUtilTest.java
+++ b/jkube-kit/common/src/test/java/org/eclipse/jkube/kit/common/util/Fabric8HttpUtilTest.java
@@ -13,18 +13,20 @@
  */
 package org.eclipse.jkube.kit.common.util;
 
-import io.fabric8.kubernetes.client.http.HttpResponse;
-import io.fabric8.kubernetes.client.http.TestHttpResponse;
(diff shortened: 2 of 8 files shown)


jkube-kit/common/src/main/java/org/eclipse/jkube/kit/common/util/Fabric8HttpUtil.java
```diff
@@ -71,4 +71,23 @@ public static String toFormData(Map<String, String> formData) throws Unsupported
...
+  public static String appendAdditionalScopeIfNotPresent(String scopeStr, String value) {
+    String[] scopeParts = scopeStr.split(":");
+    if (scopeParts.length > 2) {
+      String pullOrPushScope = scopeParts[2];
+      if (!pullOrPushScope.contains(value)) {
+        scopeParts[2] = scopeParts[2] + "," + value;
+      }
+    }
+    return String.join(":", scopeParts);
+  }
```
manusa, line 92: This is still not OK.

See the ref for [WWW-Authenticate](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/WWW-Authenticate)

We first need to identify the appropriate `WWW-Authenticate` header and challenge for the current realm.

Then it might be interesting to parse the challenges as a Map or more specifically as Properties and then retrieve the one that we're interested in (`scope`).

manusa: I wouldn't say that that method is parsing the header and reading properties

manusa: OK, I think I'm getting confused here.
In any case, that method might benefit from some built-in Java property parsing.
In addition, the signature for the return should probably be changed to something like:
```java
public static List<Properties> extractAuthenticationChallengeIntoMap(HttpResponse<?> response)
```

manusa: Please also check the [WWW-Authenticate](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/WWW-Authenticate) spec documentation.

Probably that's why the parsing is not being done correctly and that might be clearly the source of the problem.

jkube-kit/common/src/test/java/org/eclipse/jkube/kit/common/util/Fabric8HttpUtilTest.java
```diff
@@ -49,7 +50,91 @@ void toFormData_whenDataProvidedAsMap_thenCreateFormDataPayload() throws Unsuppo
...
+        .hasSize(2)
+        .containsEntry("Basic realm", "Access to the staging site")
+        .containsEntry("charset", "UTF-8");
+  }
+
+  @Test
+  void extractAuthenticationChallengeIntoMap_whenMultipleChallenges_thenParseDataIntoMap() throws IOException {
+    // Given
+    Map<String, List<String>> responseHeaders = new HashMap<>();
+    responseHeaders.put("WWW-Authenticate", Collections.singletonList("Newauth realm=\"apps\", type=1, title=\"Login to \"apps\"\", Basic realm=\"simple\""));
```
manusa, line 83: This is what I was referring to in the internal call.
I don't think that this is a valid value, it will probably come with escaped quotations marks, as in `title=\"something \\\"escaped\\\"\"`

The idea would be to reuse something that performs the parsing for us. Not sure if Jackson, XML, or something we can leverage

manusa: So, for example, OkHttp includes the 
HttpHeadrs.parseChallenges method

PR #2628, opened 2024-02-06
ImageName should throw exception when digest length is too small

### LINKED ISSUE #2543: ImageName should throw exception when digest length is too small
### Component

None

### Description

ImageName seems to do a very simple validation on Digest in image name:
https://github.com/eclipse/jkube/blob/b47fa4a9ab9a839bd4f8a093b4b0a4d8e9c88e9f/jkube-kit/config/image/src/main/java/org/eclipse/jkube/kit/config/image/ImageName.java#L353

This allows even image names with invalid digest like `repo@sha256:ffffffffffffffffffffffffffffffffff` to pass validation check.

We need to update the Digest regexp to match what's used in [Docker Distribution](https://github.com/distribution/reference/blob/8507c7fcf0da9f570540c958ea7b972c30eeaeca/regexp.go#L81C15-L81C86):
```
[A-Za-z][A-Za-z0-9]*(?:[-_+.][A-Za-z][A-Za-z0-9]*)*[:][[:xdigit:]]{32,}
```

### Expected Behavior
ImageName digest regexp should be more strict

### Acceptance Criteria
- [ ] ImageName's digest regexp is updated with what's in Docker Distribution

Devashishbasu:
Fixes #2543 

## Type of change

 - [ x] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/jkube-kit/config/image/src/main/java/org/eclipse/jkube/kit/config/image/ImageName.java
+++ b/jkube-kit/config/image/src/main/java/org/eclipse/jkube/kit/config/image/ImageName.java
@@ -357,5 +357,5 @@ public class ImageName {
     // https://github.com/docker/docker/blob/04da4041757370fb6f85510c8977c5a18ddae380/vendor/github.com/docker/distribution/reference/regexp.go#L37
     private static final Pattern TAG_REGEXP = Pattern.compile("^[\\w][\\w.-]{0,127}$");
 
-    private static final Pattern DIGEST_REGEXP = Pattern.compile("^sha256:[a-z0-9]{32,}$");
+    private static final Pattern DIGEST_REGEXP = Pattern.compile("^[A-Za-z][A-Za-z0-9]*(?:[-_+.][A-Za-z][A-Za-z0-9]*)*[:][[:xdigit:]]{32,}$");
 }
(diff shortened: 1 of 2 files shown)


manusa:
### Eclipse JKube [CI Report](https://github.com/jkubeio/ci)

Started new GH workflow run for https://github.com/eclipse/jkube/pull/2628 (_2024-02-06T18:58:59Z_)

:gear: [JKube E2E Tests (7804713386)](https://github.com/jkubeio/ci/actions/runs/7804713386) :hourglass_flowing_sand:
- :hourglass_flowing_sand: [Bootstrap CI test run (#2628)](https://github.com/jkubeio/ci/actions/runs/7804713386/job/21287307263)

Devashishbasu:
Hi @rohanKanojia i have update digest regexp to align with whats in docker's distribution ....but some test cases are failing for example:-
java.lang.IllegalArgumentException: Given Docker name 'test:5000/repo@sha256:ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff' is invalid:
   * digest part 'sha256:ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff' doesn't match allowed pattern '^[A-Za-z][A-Za-z0-9]*(?:[-_+.][A-Za-z][A-Za-z0-9]*)*[:][[:xdigit:]]{32,}$'
See http://bit.ly/docker_image_fmt for more details

'localhost/hello-world@sha256:b94d27b9934d3e08a52e52d7da7dabfac484efe37a5380ee9088f7ace2efcde9' is invalid:
   * digest part 'sha256:b94d27b9934d3e08a52e52d7da7dabfac484efe37a5380ee9088f7ace2efcde9' doesn't match allowed pattern '^[A-Za-z][A-Za-z0-9]*(?:[-_+.][A-Za-z][A-Za-z0-9]*)*[:][[:xdigit:]]{32,}$'

As discussed with you ..we need to align the regex with docker's distribution ...then i think we need to modify the test cases according to regex whats in docker's distribution....because the test cases which are failing might be were according to old digest regexp 
private static final Pattern DIGEST_REGEXP = Pattern.compile("^sha256:[a-z0-9]{32,}$");

Devashishbasu:
@rohanKanojia It seems like the Java method we are using expects only the digest part for validation, while the provided Docker name includes additional information such as the image name and tag ..when I remove the additional information and keep only the digest part then the test cases are passing

Devashishbasu:
> @rohanKanojia It seems like the Java method we are using expects only the digest part for validation, while the provided Docker name includes additional information such as the image name and tag ..when I remove the additional information and keep only the digest part then the test cases are passing

could you please guide me what to do next?

rohanKanojia:
> As discussed with you ..we need to align the regex with docker's distribution ...then i think we need to modify the test cases according to regex whats in docker's distribution....because the test cases which are failing might be were according to old digest regexp

that's quite strange, I had ported this test from Docker Distribution source, see [reference_test.go](https://github.com/distribution/reference/blob/main/reference_test.go#L63) . Need to debug why it's not working as expected.

PR #2772, opened 2024-03-06
Add and fix warnings for various Helm goals

Jurrie:
## Description
Add and fix warnings for various Helm goals

- helm-lint and helm-push: Add warning when 'helm' goal is not ran
- helm: Fix warning when 'resource' goal is not ran

## Type of change

## Checklist
 - [ ] I have read the [contributing guidelines](https://www.eclipse.dev/jkube/contributing)
 - [ ] I signed-off my commit with a user that has signed the [Eclipse Contributor Agreement](https://www.eclipse.org/legal/ECA.php)
 - [ ] My code follows the style guidelines of this project
 - [ ] I Keep It Small and Simple: The smaller the PR is, the easier it is to review and have it merged
 - [ ] I use [conventional commits](https://www.conventionalcommits.org/) in my commit messages
 - [ ] I have performed a self-review of my code

### DIFF
--- a/jkube-kit/parent/pom.xml
+++ b/jkube-kit/parent/pom.xml
@@ -736,6 +736,12 @@
         <version>${version.mockito}</version>
         <scope>test</scope>
       </dependency>
+      <dependency>
+        <groupId>org.mockito</groupId>
+        <artifactId>mockito-inline</artifactId>
+        <version>${version.mockito}</version>
+        <scope>test</scope>
+      </dependency>
 
       <dependency>
         <groupId>io.fabric8</groupId>
--- a/kubernetes-maven-plugin/plugin/src/main/java/org/eclipse/jkube/maven/plugin/mojo/build/HelmLintMojo.java
+++ b/kubernetes-maven-plugin/plugin/src/main/java/org/eclipse/jkube/maven/plugin/mojo/build/HelmLintMojo.java
@@ -13,16 +13,43 @@
  */
 package org.eclipse.jkube.maven.plugin.mojo.build;
 
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
 import org.apache.maven.plugin.MojoExecutionException;
+import org.apache.maven.plugin.MojoFailureException;
 import org.apache.maven.plugins.annotations.LifecyclePhase;
 import org.apache.maven.plugins.annotations.Mojo;
 import org.apache.maven.plugins.annotations.ResolutionScope;
+import org.eclipse.jkube.kit.resource.helm.HelmConfig;
+import org.eclipse.jkube.kit.resource.helm.HelmService;
 
 @Mojo(name = "helm-lint", defaultPhase = LifecyclePhase.INTEGRATION_TEST, requiresDependencyResolution = ResolutionScope.COMPILE)
 public class HelmLintMojo extends AbstractHelmMojo {
 
   @Override
+  public void init() throws MojoFailureException {
+    super.init();
+
+    checkChartsExist(getHelm());
+  }
+
+  private void checkChartsExist(final HelmConfig helmConfig) {
+    for (HelmConfig.HelmType helmType : helmConfig.getTypes()) {
+      final Path chart = Paths.get(helmConfig.getOutputDir(), helmType.getOutputDir(), HelmService.CHART_FILENAME);
+      if (Files.notExists(chart)) {
+        logChartNotFoundWarning(chart);
+      }
+    }
+  }
+
+  @Override
   public void executeInternal() throws MojoExecutionException {
     jkubeServiceHub.getHelmService().lint(getHelm());
+  }
+
+  protected void logChartNotFoundWarning(final Path chart) {
+    getKitLogger().warn("No Helm chart has been generated yet by the k8s:helm goal at: " + chart);
   }
 }
--- a/jkube-kit/helm/src/main/java/org/eclipse/jkube/kit/resource/helm/HelmServiceUtil.java
+++ b/jkube-kit/helm/src/main/java/org/eclipse/jkube/kit/resource/helm/HelmServiceUtil.java
@@ -108,7 +108,7 @@ public class HelmServiceUtil {
     helmConfig.setSourceDir(resolveFromPropertyOrDefault(PROPERTY_SOURCE_DIR, project, helmConfig::getSourceDir,
         () -> separatorsToSystem(String.format("%s/META-INF/jkube/", project.getOutputDirectory()))));
     helmConfig.setOutputDir(resolveFromPropertyOrDefault(PROPERTY_OUTPUT_DIR, project, helmConfig::getOutputDir,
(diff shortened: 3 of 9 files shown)


manusa:
### Eclipse JKube [CI Report](https://github.com/jkubeio/ci)

Started new GH workflow run for https://github.com/eclipse/jkube/pull/2772 (_2024-03-06T12:03:34Z_)

:gear: [JKube E2E Tests (8171712528)](https://github.com/jkubeio/ci/actions/runs/8171712528) :hourglass_flowing_sand:
- :hourglass_flowing_sand: [Bootstrap CI test run (#2772)](https://github.com/jkubeio/ci/actions/runs/8171712528/job/22340541476)

jkube-kit/parent/pom.xml
```diff
@@ -736,6 +736,12 @@
         <version>${version.mockito}</version>
         <scope>test</scope>
       </dependency>
+      <dependency>
+        <groupId>org.mockito</groupId>
+        <artifactId>mockito-inline</artifactId>
```
Jurrie, line 741: Very strange that I needed this in OpenshiftHelmPushMojoTest for the `org.mockito.Mockito.mockConstruction(HelmService.class);` thingy, but HelmPushMojoTest doesn't need it...

kubernetes-maven-plugin/plugin/src/main/java/org/eclipse/jkube/maven/plugin/mojo/build/HelmLintMojo.java
```diff
@@ -13,16 +13,43 @@
...
+    }
+  }
+
   @Override
   public void executeInternal() throws MojoExecutionException {
     jkubeServiceHub.getHelmService().lint(getHelm());
   }
+
+  protected void logChartNotFoundWarning(final Path chart) {
+    getKitLogger().warn("No Helm chart has been generated yet by the k8s:helm goal at: " + chart);
```
rohanKanojia, line 53: Can we use `getLogPrefix()` here so that we don't need to override method in OpenShift Maven Plugin just for changing prefix?

Jurrie: We could, as `getLogPrefix()` would return the same as the goal prefix. But those two are not related.

The `getLogPrefix()` for "k8s" is defined in [`org.eclipse.jkube.maven.plugin.mojo.build.AbstractJKubeMojo.DEFAULT_LOG_PREFIX`](https://github.com/Jurrie/jkube/blob/ab0e10361a71635116eaa7c5db5a8ddf07745464/kubernetes-maven-plugin/plugin/src/main/java/org/eclipse/jkube/maven/plugin/mojo/build/AbstractJKubeMojo.java#L62)) while the mojo prefix is defined in the [pom.xml](https://github.com/Jurrie/jkube/blob/ab0e10361a71635116eaa7c5db5a8ddf07745464/kubernetes-maven-plugin/plugin/pom.xml#L181). There is nothing that relates the two, other than the values being the same.

Same goes for "oc": the log prefix is defined in [`org.eclipse.jkube.maven.plugin.mojo.OpenShift.DEFAULT_LOG_PREFIX`](https://github.com/Jurrie/jkube/blob/ab0e10361a71635116eaa7c5db5a8ddf07745464/openshift-maven-plugin/plugin/src/main/java/org/eclipse/jkube/maven/plugin/mojo/OpenShift.java#L23) and the mojo prefix in the [pom.xml](https://github.com/Jurrie/jkube/blob/ab0e10361a71635116eaa7c5db5a8ddf07745464/openshift-maven-plugin/plugin/pom.xml#L70).

Because nothing is guaranteeing that those values are (and stay) the same, I did not use `getLogPrefix()`.

rohanKanojia: > Because nothing is guaranteeing that those values are (and stay) the same

Sorry, I didn't understand this correctly. Could you please elaborate on this?

Jurrie: For example: I can change [the logger prefix](https://github.com/Jurrie/jkube/blob/ab0e10361a71635116eaa7c5db5a8ddf07745464/openshift-maven-plugin/plugin/src/main/java/org/eclipse/jkube/maven/plugin/mojo/OpenShift.java#L23) to read "ocp" instead of "oc". Then the Maven goal would still be "oc:helm"; it would not change to "ocp:helm". Also no error would occur, and no unit test would fail. That's what I mean with "nothing is guaranteeing that those values are (and stay) the same."

Because of this, you can not assume `getLogPrefix()` will return the same value as the Maven mojo prefix.

rohanKanojia: Thanks for clarifying!

rohanKanojia:
Why have you deleted these tests? Were they breaking due to this refactor?

Jurrie:
> Why have you deleted these tests? Were they breaking due to this refactor?

Yes, due to the new check in [HelmServiceUtil](https://github.com/eclipse/jkube/pull/2772/files#diff-69310bc659bfce2dd55933e303523a9b0e9d168b4df60b7d4f7b1e88a663635fR244). This check results in no exception being thrown.

The way I look at the deleted unit tests is that they were made _after_ the code was made. It more or less just models what the code was doing, instead of what the intent was. I don't see why an exception should be thrown when there are no source files. And _if_ there is a reason for it, why wasn't a dedicated exception, with explanatory name and proper description string used?

But I could be wrong here! If the tests served a purpose, please describe that purpose and I'll see what I can do to put the tests back.

jkube-kit/parent/pom.xml
rohanKanojia: I think you can avoid this by adding [`mockito-inline`](https://github.com/eclipse/jkube/tree/master/kubernetes-maven-plugin/plugin/src/test/resources/mockito-extensions) folder in `src/test/resources`

Jurrie: Wauw, I never knew! Thanks for the tip! Implemented in e463ef777222592ff93e4520d555c38d92b24e25.

manusa:
Hi Jurrie, thanks for looking into this.

This is fine but it will need to be immediately refactored if merged as is.
The idea is that this is implemented at the service level and not at the Mojo/plugin level, this way, the behavior will be consistent regardless of how the Helm feature is triggered.

One approach can be to provide the necessary variables such as the kind of plugin (Maven or Gradle) and applicable name of the goal/task (`k8s:resource`, `k8sResource`, `k8s:helm`, `k8sHelm`) to be able to compute the appropriate exception and helper message.)
Other can be to craft a specific Exception (preferably unchecked) and then compute the message at the plugin level.
Or any other thing that you can think of.

PR #2802, opened 2024-03-11
refactor (jkube-kit) : Move Dockerfile opinionated ImageConfig logic from `ConfigHelper.initImageConfiguration` to a dedicated generator

rohanKanojia:
## Description

+ ConfigHelper modifies the provided set of ImageConfigurations when it detects a `Dockerfile` in project root directory. This is actually a feature called Simple Dockerfile Mode. However, this mutation of ImageConfiguration from within `initImageConfiguration` does not look appropriate. It can be moved to a dedicated generator that will get activated when `Dockerfile` is present in project base directory.
+ Add a new generator SimpleDockerfileGenerator, it adds a simple ImageConfiguration with Dockerfile configured in build and adds it to existing set of ImageConfigurations.
+ Remove check from BaseGenerator to skip generation if it's simple Dockerfile mode. Move it to SimpleDockerfileGenerator.isApplicable instead

We also need to adjust the documentation for adding this new generator. I'll do that in a follow up PR.

## Type of change

## Checklist
 - [x] I tested my code in Kubernetes


manusa:
I recall we might have discussed this a few weeks ago.
We need to ensure we have enough tests that guarantee that in case the order of generators is altered it gets reported.
This idea is really good to clean out the code but it has the fragility of depending on the generator order.

rohanKanojia:
@manusa : I had added a test in SimpleDockerFileGeneratorTest to verify that generator doesn't modify existing ImageConfiguration when they have been altered by any other generator.

Shall I add an additional test to assert the order of generators in the default profile?

rohanKanojia:
I've added tests verifying the generators execution order in the default profile in b5d9de6753534bc9b94f663d5c3941e83ac1b188 . Please check if it's in line with your expectations.

jkube-kit/generator/dockerfile-simple/src/main/java/org/eclipse/jkube/generator/dockerfile/simple/SimpleDockerfileGenerator.java
```diff
@@ -0,0 +1,57 @@
...
+import static org.eclipse.jkube.kit.common.util.PropertiesUtil.getValueFromProperties;
+
+public class SimpleDockerfileGenerator extends BaseGenerator {
+  public SimpleDockerfileGenerator(GeneratorContext context) {
+    super(context, "dockerfile-simple");
+  }
+
+  @Override
+  public boolean isApplicable(List<ImageConfiguration> configs) {
+    return shouldAddGeneratedImageConfiguration(configs) &&
```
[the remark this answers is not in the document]
rohanKanojia: You're right. Thanks a lot for noticing. Let me update it.

gradle-plugin/kubernetes/src/test/java/org/eclipse/jkube/gradle/plugin/task/KubernetesBuildTaskTest.java
```diff
@@ -137,4 +159,43 @@ public Property<Boolean> getSkipBuild() {
...
+        .properties(new Properties()).build();
+    when(taskEnvironment.project.getExtensions().getByType(KubernetesExtension.class)).thenReturn(extension);
+    final KubernetesBuildTask buildTask = new KubernetesBuildTask(KubernetesExtension.class);
+    GeneratorContext generatorContext = buildTask.initGeneratorContextBuilder().build();
+
+    // When
+    List<Generator> generatorList = createUsableGeneratorList(generatorContext);
+
+    // Then
+    Assertions.assertThat(generatorList)
```
sunix, line 186: for readability purpose, could you make Assertions.assertThat statically imported ?

jkube-kit/generator/dockerfile-simple/pom.xml
```diff
@@ -0,0 +1,47 @@
...
+  <dependencies>
+    <dependency>
+      <groupId>org.eclipse.jkube</groupId>
+      <artifactId>jkube-kit-generator-api</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter-engine</artifactId>
+      <scope>test</scope>
```
sunix, line 40: I would keep the scope declaration consistent:
- either you don't declare them and assume it is coming from the dependencyManagement
- either you declare them everywhere (also for org.assertj)

kubernetes-maven-plugin/plugin/src/test/java/org/eclipse/jkube/maven/plugin/mojo/develop/WatchMojoTest.java
```diff
@@ -84,6 +84,7 @@ void setUp(@TempDir Path temporaryFolder) throws Exception {
     when(mavenProject.getCompileClasspathElements()).thenReturn(Collections.emptyList());
     when(mavenProject.getDependencies()).thenReturn(Collections.emptyList());
     when(mavenProject.getDevelopers()).thenReturn(Collections.emptyList());
+    when(mavenProject.getBasedir()).thenReturn(temporaryFolder.toFile());
```
sunix, line 87: is this related to refactor ?

rohanKanojia: `SimpleDockerfileGenerator.isApplicable` method requires project base directory to be provided. `mavenProject` is mocked with deep stubs; when we invoke `mavenProject.getBasedir()` it is returning some mocked object which is causing NPE when this base directory is checked in  `SimpleDockerfileGenerator.isApplicable`

gradle-plugin/kubernetes/src/test/java/org/eclipse/jkube/gradle/plugin/task/KubernetesBuildTaskTest.java
```diff
@@ -137,4 +159,43 @@ public Property<Boolean> getSkipBuild() {
...
+    final KubernetesBuildTask buildTask = new KubernetesBuildTask(KubernetesExtension.class);
+    GeneratorContext generatorContext = buildTask.initGeneratorContextBuilder().build();
+
+    // When
+    List<Generator> generatorList = createUsableGeneratorList(generatorContext);
+
+    // Then
+    assertThat(generatorList)
+        .hasSize(12)
+        .satisfies(g -> Assertions.assertThat(g.get(0)).isInstanceOf(SimpleDockerfileGenerator.class))
```
sunix, line 188: Same here, would be nice to have static imports for Assertions.assertThat

sunix (approved):

gradle-plugin/kubernetes/src/test/java/org/eclipse/jkube/gradle/plugin/task/KubernetesBuildTaskTest.java
rohanKanojia: Sorry, let me fix it.

manusa:
> I've added tests verifying the generators execution order in the default profile in [b5d9de6](https://github.com/eclipse/jkube/commit/b5d9de6753534bc9b94f663d5c3941e83ac1b188) . Please check if it's in line with your expectations.

This works, however is extremely brittle and hard to maintain (which probably defeats the purpose). The idea is to create a test that ensures the behavior (not sure how since I haven't dug into the code), test should fail in the event of a modification in the generator order (default profile).

rohanKanojia:
@manusa There is a `log.verbose` statement in DefaultGeneratorManager that prints the generators being invoked: 

https://github.com/eclipse/jkube/blob/4cfb00728d455273827d0552376f304104fbed24/jkube-kit/generator/api/src/main/java/org/eclipse/jkube/generator/api/DefaultGeneratorManager.java#L54-L56

Shall I try adding log verifications for this instead?

manusa:
> Shall I try adding log verifications for this instead?

If the check consists in checking the generator print order, then that's mostly the same of what we have now.
We either need to ensure that this generator gets printed the very first always, or that its behavior is consistent.
When I talk about behavior is having a test that would fail if the order of generators was altered (maybe we already have one like this, but in that case we should state explicitly or add a case where this is reported)

kubernetes-maven-plugin/plugin/src/test/java/org/eclipse/jkube/maven/plugin/mojo/build/BuildMojoTest.java
```diff
@@ -58,4 +73,35 @@ void buildServiceConfigBuilder_shouldReturnNonNullResourceConfigIfConfigured() {
             .returns("200m", c -> c.getLimits().get("cpu"))
             .returns("1Gi", c -> c.getRequests().get("memory"));
     }
+
+    @Test
+    @DisplayName("when Dockerfile present in project base directory, then generate opinionated ImageConfiguration for Dockerfile")
+    void simpleDockerfileModeWorksAsExpected(@TempDir File temporaryFolder) throws IOException {
```
manusa, line 79: These tests (task + Mojo) work, but they aren't exactly where they should actually be.
I'll try to give take it for a spin and see if we can make this easier.

manusa (approved):
LGTM, thx!

Issue #2904, opened 2024-04-12
docker.buildArg.* properties not taken into account in OpenShift plugins


sunix:
### Describe the bug

The JKube Openshift Maven plugin fails to recognize specified Docker build-arg when they are injected into a Dockerfile from Maven properties. This issue manifests specifically when using Maven properties and does not occur with the XML/Gradle DSL configuration since https://github.com/eclipse/jkube/pull/2895.
This issue occurs despite the JKube [documentation](https://eclipse.dev/jkube/docs/openshift-maven-plugin/#_build_args) confirming that build arguments can be specified either via the plugin configuration or as Maven properties.

### Eclipse JKube version

SNAPSHOT

### Component

OpenShift Maven Plugin

### Apache Maven version

None

### Gradle version

None

### Steps to reproduce

1. **Set Up the Dockerfile:**
   Place the following content in `src/main/docker/Dockerfile`:
   ```dockerfile
   ARG FILE_EXTENSION
   FROM docker.io/alpine:latest as builder
   RUN touch test.${FILE_EXTENSION}

   FROM docker.io/alpine:latest
   COPY --from=builder /test.txt /test.txt

   ENTRYPOINT ["ls"]
   ```

2. **Configure the JKube Plugin:**
   Insert the JKube Openshift Maven plugin configuration into your `pom.xml`:
   ```xml
   <plugin>
       <groupId>org.eclipse.jkube</groupId>
       <artifactId>openshift-maven-plugin</artifactId>
       <version>1.17-SNAPSHOT</version>
       <configuration>
         <buildStrategy>docker</buildStrategy>
         <images>
           <image>
             <name>helloworld:%l</name>
             <build>
               <contextDir>${project.basedir}/src/main/docker/</contextDir>
               <dockerFile>Dockerfile</dockerFile>
             </build>
           </image>
         </images>
       </configuration>
   </plugin>
   ```

3. **Define the Build Argument:**
   Define the build argument for the Docker image in your `pom.xml`:
   ```xml
   <properties>
       <docker.buildArg.FILE_EXTENSION>txt</docker.buildArg.FILE_EXTENSION>
   </properties>
   ```

4. **Execute the Build Command:**
   Execute the build using the following Maven command:
   ```
   mvn oc:build
   ```

5. **Error Observation:**
   The build process fails, yielding the following output:
   ```
   [INFO] oc: --> abd4c9e7124f
[...]

PR #2994, opened 2024-04-29
refactor (jkube-kit/build) : Move buildArgs related logic to a new class BuildArgResolverUtil

rohanKanojia:
## Description

Related to #2904

It's better to move code for resolving build args from different sources to a dedicated class. It doesn't look like BuildService is the right place for this. This way we can also use it from other BuildServices of other build strategies.

This PR just moves BuildArg related logic to a new utility class. There is one additional test for adding build args from docker proxy config.

## Type of change

## Checklist


manusa:
### Eclipse JKube [CI Report](https://github.com/eclipse-jkube/ci)

Started new GH workflow run for https://github.com/eclipse-jkube/jkube/pull/2994 (_2024-04-29T10:24:44Z_)

:gear: [JKube E2E Tests (8877028632)](https://github.com/eclipse-jkube/ci/actions/runs/8877028632) :hourglass_flowing_sand:
- :hourglass_flowing_sand: [Bootstrap CI test run (#2994)](https://github.com/eclipse-jkube/ci/actions/runs/8877028632/job/24369763055)

jkube-kit/build/api/src/main/java/org/eclipse/jkube/kit/build/api/helper/BuildArgResolverUtil.java
```diff
@@ -0,0 +1,102 @@
...
+import java.util.Properties;
+
+public class BuildArgResolverUtil {
+  private static final String ARG_PREFIX = "docker.buildArg.";
+
+  private BuildArgResolverUtil() { }
+
+  public static Map<String, String> mergeBuildArgs(ImageConfiguration imageConfig, JKubeConfiguration configuration) {
+    return prepareBuildArgs(addBuildArgs(configuration), imageConfig.getBuildConfiguration());
+  }
```
manusa, line 35: Could we use this PR to provide some sort of order and orchestration for this method instead of just chaining methods.

It'd be nice if the method would show how the args are computed and populated based on their source. This would also show how the precedence works.

```
final Map<String, String> args = new HashMap<>();
// Least precedence args added to the map
// Next args added to the map, overriding any previous entries
// ...
return args;
```

rohanKanojia: While addressing this, I noticed that we're using [Guava's  `com.google.common.collect.ImmutableMap`](https://guava.dev/releases/23.0/api/docs/com/google/common/collect/ImmutableMap.html) that throws exception when colliding keys are added.

Shall I remove its usage? At the moment, we are throwing exception on collision.

manusa: So at the moment it's not possible to override args? that's weird.
We should definitely remove Guava.
If the current behavior is that override is not possible, we can preserve it. We should also create a new issue to see if we want to change this. IMO users my want to be able to provide one-off overrides, so maybe a warning instead of an exception is better.

rohanKanojia: I've created https://github.com/eclipse-jkube/jkube/issues/3000 for this task.

manusa (changes requested):
https://github.com/rohanKanojia/jkube/pull/122

manusa (approved):
LGTM, thx!

Issue #2663, opened 2024-02-13
Helm: helm-install task/goal


manusa:
# Helm: helm-install task/goal

https://github.com/eclipse/jkube/issues/2663

## Description

As an Eclipse JKube user, I want to be able to install the chart JKube generates, so that I can deploy it to my Kubernetes/OpenShift cluster.

### JKube Kit

JKube Kit exposes a new method `install` that installs the **exploded** chart available in the coordinates (output directory, helm type, and so on -see `resolveTarballFile`-) provided in the `HelmConfig` argument.

The chart should have been previously generated using the `generateHelmCharts` method.

- The installation is performed by executing [helm-java `install` command](https://github.com/manusa/helm-java#install).
- The user is able to provide the following installation options through the HelmConfig:
  - `releaseName`: `jkube.helm.release.name`<br/>
    Optional, if not specified, the release name should be inferred following the same procedure used to compute default image and resource names.
  - `installDependencyUpdate`: `jkube.helm.install.dependencyUpdate`<br/>
    Optional, if not provided should default to `true`.<br/>
    https://github.com/eclipse/jkube/issues/2110
  - `installWaitReady`: `jkube.helm.install.waitReady`
- The installation result is logged, fields provided include:
  - `name`
  - `namespace`
  - `status`
  - `revision`
  - `lastDeployed`
- If installation fails, error message is logged

### Maven plugins

The Kubernetes Maven Plugin and OpenShift Maven Plugin expose a new `helm-install` goal in a `HelmInstallMojo` and `OpenshiftHelmInstallMojo` that extends the basic `HelmMojo` (following a similar approach to that of `HelmPushMojo` and `HelmLintMojo`).

The mojos are bound by default to the `LifecyclePhase.INSTALL` phase.

Tests should be added for each of the mojos and also for `KubernetesPluginTest`, `OpenShiftPluginTest`, and `GeneratedPluginDescriptorTest`.

### Gradle plugins

The Kubernetes Gradle Plugin and OpenShift Gradle Plugin expose a new `helmInstall` task in a `KubernetesHelmInstallTask` and `OpenShiftHelmInstallTask` that extends the basic `AbstractJKubeTask` task.

Individual tests should be added for the new tasks and also for the plugins to verify task precedence.

An [e2e test](https://github.com/jkubeio/jkube-integration-tests) proves that the task can be executed both, with a successful and a failed installation.

### Documentation

[...]

PR #3142, opened 2024-06-06
feat (jkube-kit/helm) : HelmService exposes helm install functionality

rohanKanojia:
## Description
Related to #2663

+ Add new method `prepareMockWebServerExpectationsForAggregatedDiscoveryEndpoints` in KubernetesMockServerUtil for adding manual expectations for Kubernetes Aggregated Discovery endpoints
+ Added a new `install` method in HelmService that would perform installation of a helm chart onto Kubernetes cluster using Helm  java library
+ Added configuration options in HelmConfig
   - `releaseName`
   - `installDependencyUpdate`
   - `installWaitReady`
   - `disableOpenAPIValidation`

## Type of change

## Checklist

### DIFF
--- a/jkube-kit/helm/src/main/java/org/eclipse/jkube/kit/resource/helm/HelmService.java
+++ b/jkube-kit/helm/src/main/java/org/eclipse/jkube/kit/resource/helm/HelmService.java
@@ -18,6 +18,7 @@ import java.io.IOException;
 import java.nio.charset.Charset;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -35,8 +36,12 @@ import java.util.stream.Stream;
 
 import com.marcnuri.helm.DependencyCommand;
 import com.marcnuri.helm.Helm;
+import com.marcnuri.helm.InstallCommand;
 import com.marcnuri.helm.LintCommand;
 import com.marcnuri.helm.LintResult;
+import com.marcnuri.helm.Release;
+import io.fabric8.kubernetes.client.KubernetesClient;
+import io.fabric8.kubernetes.client.KubernetesClientBuilder;
 import org.eclipse.jkube.kit.common.JKubeConfiguration;
 import org.eclipse.jkube.kit.common.JKubeException;
 import org.eclipse.jkube.kit.common.KitLogger;
@@ -63,6 +68,7 @@ import org.apache.commons.lang3.StringUtils;
 import static org.apache.commons.lang3.StringUtils.EMPTY;
 import static org.eclipse.jkube.kit.common.JKubeFileInterpolator.DEFAULT_FILTER;
 import static org.eclipse.jkube.kit.common.JKubeFileInterpolator.interpolate;
+import static org.eclipse.jkube.kit.common.util.KubernetesHelper.exportKubernetesClientConfigToFile;
 import static org.eclipse.jkube.kit.common.util.MapUtil.getNestedMap;
 import static org.eclipse.jkube.kit.common.util.TemplateUtil.escapeYamlTemplate;
 import static org.eclipse.jkube.kit.common.util.YamlUtil.listYamls;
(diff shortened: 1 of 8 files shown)


manusa:
### Eclipse JKube [CI Report](https://github.com/eclipse-jkube/ci)

Started new GH workflow run for https://github.com/eclipse-jkube/jkube/pull/3142 (_2024-06-06T14:13:23Z_)

:gear: [JKube E2E Tests (9402470308)](https://github.com/eclipse-jkube/ci/actions/runs/9402470308) :hourglass_flowing_sand:
- :hourglass_flowing_sand: [Bootstrap CI test run (#3142)](https://github.com/eclipse-jkube/ci/actions/runs/9402470308/job/25896678464)

jkube-kit/helm/src/main/java/org/eclipse/jkube/kit/resource/helm/HelmServiceUtil.java
```diff
@@ -80,6 +82,10 @@ public class HelmServiceUtil {
   protected static final String PROPERTY_HELM_DEBUG = "jkube.helm.debug";
   protected static final String PROPERTY_HELM_DEPENDENCY_VERIFY = "jkube.helm.dependencyVerify";
   protected static final String PROPERTY_HELM_DEPENDENCY_SKIP_REFRESH = "jkube.helm.dependencySkipRefresh";
+  protected static final String PROPERTY_HELM_RELEASE_NAME = "jkube.helm.release.name";
+  protected static final String PROPERTY_HELM_INSTALL_DEPENDENCY_UPDATE = "jkube.helm.install.dependencyUpdate";
+  protected static final String PROPERTY_HELM_INSTALL_WAIT_READY = "jkube.helm.install.waitReady";
+  protected static final String PROPERTY_HELM_INSTALL_DISABLE_OPENAPI_VALIDATION = "jkube.helm.disableOpenAPIValidation";
```
manusa, line 88: This doesn't follow the pattern of other settings.
Is it a generic setting (i.e. it applies to more than one helm goal -install, uninstall, upgrade-)?
Or is it a setting specific to install?

I'd say it's the former, so in this case maybe the field name is not ok.

rohanKanojia: I saw in https://github.com/manusa/helm-java/pull/133 that this field is applicable for install and update, that's why I kept the name as  global one.

I'll rename the field name.

jkube-kit/helm/src/main/java/org/eclipse/jkube/kit/resource/helm/HelmService.java
```diff
@@ -189,6 +202,56 @@ public void dependencyUpdate(HelmConfig helmConfig) {
...
+        .ifPresent(s -> Arrays.stream(s.split("\r?\n"))
+          .filter(StringUtils::isNotBlank)
+          .forEach(l -> logger.info("[[W]]%s", l)));
+    }
+  }
+
+  private Path createTemporaryKubeConfigForInstall() {
+    if (kubernetesClient == null) {
+      kubernetesClient = new ClusterAccess(ClusterConfiguration.from(
+        System.getProperties(), jKubeConfiguration.getProject().getProperties()).build()).createDefaultClient();
```
rohanKanojia, line 241: This is done to avoid other Helm goals requiring cluster access. When I was trying to initialize KubernetesClient in HelmService construction in JKubeServiceHub, it caused other Helm actions to start throwing exceptions in offline mode

jkube-kit/helm/src/main/java/org/eclipse/jkube/kit/resource/helm/HelmService.java
```diff
@@ -85,11 +92,17 @@ public class HelmService {
   private final JKubeConfiguration jKubeConfiguration;
   private final ResourceServiceConfig resourceServiceConfig;
   private final KitLogger logger;
+  private KubernetesClient kubernetesClient;
 
   public HelmService(JKubeConfiguration jKubeConfiguration, ResourceServiceConfig resourceServiceConfig, KitLogger logger) {
+    this(jKubeConfiguration, resourceServiceConfig, logger, null);
+  }
+
+  HelmService(JKubeConfiguration jKubeConfiguration, ResourceServiceConfig resourceServiceConfig, KitLogger logger, KubernetesClient kubernetesClient) {
```
manusa, line 101: I'm reviewing and thinking about this PR since yesterday.

The fact that we're adding extra fields to this constructor doesn't feel right (especially one that's not a configuration but a heavy KubernetesClient instance).

We should probably be able to infer the configuration using `ClusterConfiguration.from(access, System.getProperties(), project.getProperties()).build();`

just like we do here:

https://github.com/eclipse-jkube/jkube/blob/ad4233e5b4fc0d6455481fec77ccfd7b510a8e05/jkube-kit/config/service/src/main/java/org/eclipse/jkube/kit/config/service/JKubeServiceHub.java#L128-L137

And then use the ClusterAccess#getConfig to initialize the temporary `.kube/config`

Maybe we can discuss this later, to see if there are better alternatives.
But storing an KubernetesClient instance here, doesn't seem right at all.

manusa (changes requested):

jkube-kit/helm/src/main/java/org/eclipse/jkube/kit/resource/helm/HelmService.java
rohanKanojia: I had added this mainly for testing. Exposing KubernetesClient field from constructor was causing other helm features to depend on K8s. It's only package private constructor, this KubernetesClient field would not be exposed via public constructor (it's passed as `null`)

manusa: It doesn't matter, it's still not right to do so.
The helm module has no beef with the Kubernetes Client, it only has the dependency because the kubernetes-client dependency is high up in the hierarchy.
Let's say we want to split this in the future to allow for a helm-maven-plugin, the only dependency we'd probably want at that point would be the helm-java one, nothing else.

manusa (approved):
LGTM, thx!

PR #3980, opened 2026-08-19
fix(spring-boot): use tools jarmode for layer extraction on Spring Boot 4.1+ (3978)

### LINKED ISSUE #3978: Spring Boot generator fails with Unsupported jarmode 'layertools' on Spring Boot 4.1+ (jarmode renamed to tools)
### Describe the bug

The Spring Boot generator's layer extraction (SpringBootLayeredJar.extractLayers()) hardcodes:

`return ArrayUtils.addAll(new String[] { "java", "-Djarmode=layertools", "-jar", layeredJar.getAbsolutePath()}, args);`

https://github.com/eclipse-jkube/jkube/blob/master/jkube-kit/jkube-kit-spring-boot/src/main/java/org/eclipse/jkube/springboot/SpringBootLayeredJar.java#L103

Spring Boot 4.1.0 removed the layertools jarmode entirely and replaced it with the unified tools jarmode (tracked in spring-projects/spring-boot#48568, milestone 4.1.0-M1, closed). A jar built with Boot 4.1 no longer registers a layertools handler, so this call now always fails:

>[WARNING] k8s: Error: Unsupported jarmode 'layertools'
>[ERROR] ... Failure in extracting spring boot jar layers: Process 'java -Djarmode=layertools -jar app.jar extract' exited with status 1

This makes it impossible to build a layered container image via the spring-boot generator for any Spring Boot 4.1+ project — the build fails outright, there's no fallback to a non-layered image, and there's no config flag to change the jarmode string.

### Eclipse JKube version

1.20.0

### Component

Kubernetes Maven Plugin

### Apache Maven version

3.6.3

### Gradle version

_No response_

### Steps to reproduce

1. Create a Spring Boot 4.1.0 project with a kubernetes-maven-plugin config using the spring-boot generator (default layering enabled).
2. Run `mvn package k8s:build`.
3. Observe the failure above.

### Expected behavior

The generator extracts layers successfully (possibly using `-Djarmode=tools -jar app.jar extract --layers --destination <dir>` for Boot 4.1+, keeping `-Djarmode=layertools` for older Boot versions).

### Runtime

other (please specify in additional context)

other (please specify in additional context)

### Environment

Linux

### Eclipse JKube Logs

```shell
[INFO] k8s: Running generator spring-boot
[INFO] k8s: spring-boot: Using Docker image eclipse-temurin:25-jre as base / builder
[INFO] k8s: Spring Boot layered jar detected
[WARNING] k8s: Error: Unsupported jarmode 'layertools'
[WARNING] k8s: 
[INFO] BUILD FAILURE
[INFO] Total time:  7.608 s
[INFO] Finished at: 2026-08-18T17:07:14+02:00
[ERROR] Failed to execute goal org.eclipse.jkube:kubernetes-maven-plugin:1.20.0:build (default-cli) on project jkube-layertools-repro: Execution default-cli of goal org.eclipse.jkube:kubernetes-maven-plugin:1.20.0:build failed: Failure in extracting spring boot jar layers: Process 'java -Djarmode=layertools -jar /home/dcavestro@siav.net/tmp/jkube-layertools-repro/target/jkube-layertools-repro.jar extract' exited with status 1 -> [Help 1]
```

### pom.xml
```xml
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0"
         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
    <modelVersion>4.0.0</modelVersion>

    <groupId>com.acme</groupId>
    <artifactId>jkube-layertools-repro</artifactId>
    <version>0.0.1-SNAPSHOT</version>
    <packaging>jar</packaging>

    <properties>
        <maven.compiler.release>25</maven.compiler.release>
        <spring-boot.version>4.1.0</spring-boot.version>
    </properties>

    <dependencies>
        <dependency>
            <groupId>org.springframework.boot</groupId>
            <artifactId>spring-boot-starter</artifactId>
            <version>${spring-boot.version}</version>
        </dependency>
    </dependencies>

    <build>
        <finalName>jkube-layertools-repro</finalName>
        <plugins>
            <!-- Default layout: repackage produces a layered jar (BOOT-INF/layers.idx) by default in Boot 3+/4.x -->
            <plugin>
                <groupId>org.springframework.boot</groupId>
                <artifactId>spring-boot-maven-plugin</artifactId>
                <version>${spring-boot.version}</version>
                <configuration>
                    <mainClass>com.acme.repro.Application</mainClass>
                </configuration>
                <executions>
                    <execution>
                        <goals>
                            <goal>repackage</goal>
                        </goals>
                    </execution>
                </executions>
            </plugin>

            <!--
                Reproducer for: jkube's spring-boot generator hardcodes '-Djarmode=layertools',
                which Spring Boot 4.1 removed (replaced by the unified 'tools' jarmode).

                Run: mvn package k8s:build
                Expected (buggy) result:
                  [WARNING] k8s: Error: Unsupported jarmode 'layertools'
                  [ERROR] ... Failure in extracting spring boot jar layers: Process 'java -Djarmode=layertools -jar ... extract' exited with status 1
            -->
            <plugin>
                <version>1.20.0</version>
                <configuration>
                    <buildStrategy>jib</buildStrategy>
                    <generator>
                        <includes>
                            <include>spring-boot</include>
                        </includes>
                        <config>
                            <spring-boot>
                                <from>eclipse-temurin:25-jre</from>
                                <name>jkube-layertools-repro:latest</name>
                            </spring-boot>
                        </config>
                    </generator>
                </configuration>
            </plugin>
        </plugins>
    </build>
</project>
```
### src/main/java/com/acme/repro/Application.java
```java
package com.acme.repro;

import org.springframework.boot.SpringApplication;
import org.springframework.boot.autoconfigure.SpringBootApplication;

@SpringBootApplication
public class Application {

    public static void main(String[] args) {
        SpringApplication.run(Application.class, args);
    }
}
```

### Additional context

Environment
- jkube kubernetes-maven-plugin 1.18.1 (confirmed still present on master / v1.20.0)
- Spring Boot 4.1.0
- Build strategy: jib (also affects docker/buildpacks — the failure happens before the builder is invoked)

ash-thakur-rh:
## Overview

Fixes #3978 - Spring Boot generator fails with "Unsupported jarmode 'layertools'" on Spring Boot 4.1+

Spring Boot 4.1.0 removed the `layertools` jarmode and replaced it with the unified `tools` jarmode. This PR adds automatic version detection and proper command-line flags to support all Spring Boot versions from 2.7.x through 4.1+.

### 1. Version-Based Jarmode Selection

**Added automatic version detection:**
- Spring Boot >= 4.1.0: Uses `tools` jarmode
- Spring Boot < 4.1.0: Uses `layertools` jarmode
- Missing version: Falls back to trying both modes (tools → layertools)

**Implementation:**
- New method `isVersion410OrNewer()` to parse and compare Spring Boot versions
- `determineJarMode()` reads version from jar manifest and selects appropriate jarmode
- Returns `null` if version cannot be determined, triggering fallback mechanism

### 2. Idempotent Extraction with `--destination` and `--force`

**Fixed directory structure and idempotent builds:**
- Added `--destination .` flag: Layers now extract directly to `buildPackageDirectory` instead of `buildPackageDirectory/<jar-basename>/`
- Added `--force` flag for `tools` jarmode: Enables repeated builds without `mvn clean`
- `layertools` jarmode: Uses `--destination .` (doesn't support `--force`)

**Command-line arguments:**
```java
// tools jarmode (Spring Boot 4.1+)
["extract", "--launcher", "--layers", "--destination", ".", "--force"]

// layertools jarmode (Spring Boot < 4.1)
["extract", "--destination", "."]
```

**Benefits:**
- ✅ `mvn package k8s:build` can run multiple times without clean
- ✅ Consistent directory layout across all Spring Boot versions

### 3. Robust Fallback Mechanism

**Improved error handling:**
- Only runs fallback when version detection fails (`jarMode == null`)
- Never re-runs the mode that just failed (gated with `if (jarMode == null)`)
- Reports the **primary** exception cause, not misleading fallback errors
- Uses `addSuppressed()` to preserve full error context
- Logs success/failure at appropriate levels with jarmode information

**Example error flow:**
1. Detect version → Spring Boot 4.1.0 → try `tools` jarmode
2. If `tools` fails: throw `IllegalStateException` with the **actual** `tools` failure (not fallback errors)
3. If version missing: try `tools` → if fails, try `layertools` → report first failure as primary

### 4. Custom Layer Names Support

**No hardcoded layer assumptions:**
- Uses `listLayers()` to read actual layer names from `BOOT-INF/layers.idx`
- Supports custom layer configurations via `layers.xml` in spring-boot-maven-plugin
- Works with any layer names (e.g., `libs`, `loader`, `snapshots`, `app` instead of standard names)

### 5. Testability Improvements

**Added package-private testing seam:**
- `executeLayerToolsCommand()` allows tests to observe actual command invocations
- Tests verify real production behavior, not mocked data

**Test coverage:**
- Version parsing and comparison logic (29 test cases)
- Jarmode selection based on version (5 test cases)  
- Command-line arguments for each jarmode (4 test cases)
- Fallback sequence verification (2 test cases)
- Custom layer names support (1 test case)
- Directory path verification (1 test case)
- **Total: 281 tests** (all passing)

### 6. Code Quality Fixes

**Fixed logging issues:**
- Changed exception logging to use `e.getMessage()` instead of passing `Throwable` directly
- `KitLogger.debug()` with multiple params doesn't unwrap exceptions - now explicitly logs message

**Removed obsolete code:**
- Deleted `findLayerBaseDirectory()` method and related subdirectory search logic
- Removed 9 obsolete tests that were testing non-existent behavior
- Simplified helper methods

## Breaking Changes

None. This is fully backward compatible:
- Spring Boot < 4.1: Continues to use `layertools` jarmode
- Spring Boot >= 4.1: Automatically uses `tools` jarmode
- Existing projects will see no behavior change except fixed builds for 4.1+

## Related Issues

Closes #3978


jkube-kit/jkube-kit-spring-boot/src/main/java/org/eclipse/jkube/springboot/SpringBootLayeredJar.java
```diff
@@ -81,26 +102,92 @@ public List<String> listLayers() {
...
+    throw new IllegalStateException("Failure in extracting spring boot jar layers", lastException);
+  }
+
+  // Package-private for testing
+  String[] getExtractArgs(String jarMode) {
+    // Spring Boot 4.x tools jarmode requires --launcher and --layers flags
+    // to produce layered directory structure with spring-boot-loader classes included
+    // layertools jarmode only supports extract command without flags
+    if (JARMODE_TOOLS.equals(jarMode)) {
+      return new String[]{"extract", "--launcher", "--layers"};
```
manusa: **Missing `--destination .` and `--force` — the second build of any Spring Boot >= 3.3 project fails.**

Spring's own `layertools` compat shim (`ExtractLayersCommand`) delegates to `ExtractCommand` with exactly `--destination "." --layers --launcher --force`. This line is that command minus the last two, and both omissions bite:

- **No `--destination`** -> `ExtractCommand.getDestination()` returns `new File(workingDir, stripExtension(jarName))`, i.e. `target/<jar-basename>/` rather than `target/`. That is the only reason `findLayerBaseDirectory` below has to exist.
- **No `--force`** -> `checkDirectoryIsEmpty()` throws when the destination is a non-empty directory.

Reproduced against a real Spring Boot 4.1.1 layered jar, run through the real `JarLauncher`:

```
build1: exit=0    -> target/app-1.0.0/{dependencies,spring-boot-loader,snapshot-dependencies,application}
build2: exit=1    JarModeErrorException: .../app-1.0.0 already exists and is not empty
                  at ExtractCommand.checkDirectoryIsEmpty(ExtractCommand.java:142)
```

So `mvn package k8s:build` twice without `clean` fails. JKube then retries `tools` (fails identically) and finally `layertools`, which on 4.1+ no longer exists — three JVM forks, then a hard failure, on the very version this PR fixes.

```java
return new String[]{"extract", "--launcher", "--layers", "--destination", ".", "--force"};
```

I verified all four options are present in every `spring-boot-jarmode-tools` release from 3.3.0 to 4.1.1; that this runs clean three times in a row; and that on 3.5.0 its output tree is `diff`-identical to `layertools extract`. It also restores today's on-disk layout, which lets `findLayerBaseDirectory`/`getJarBaseName` be deleted.

jkube-kit/jkube-kit-spring-boot/src/main/java/org/eclipse/jkube/springboot/SpringBootLayeredJar.java
```diff
@@ -81,26 +102,92 @@ public List<String> listLayers() {
...
-    } catch (IOException ioException) {
-      throw new IllegalStateException("Failure in extracting spring boot jar layers", ioException);
+      String[] parts = version.split("[.-]");
+      if (parts.length < 2) {
+        return false;
+      }
+      int major = Integer.parseInt(parts[0]);
+      int minor = Integer.parseInt(parts[1]);
+
+      return major > 3 || (major == 3 && minor >= 3);
```
manusa: **The 3.3.0 threshold is wider than the bug.**

I checked `META-INF/spring.factories` in every `spring-boot-jarmode-tools` release: `LayerToolsJarMode` is registered in 3.3.0, 3.4.13, 3.5.0 and 4.0.8, and is **absent from 4.1.0 and 4.1.1** (the class is removed entirely). So `-Djarmode=layertools` still works everywhere below 4.1.

As written, this changes the command, the extracted layout and the generator's directory resolution for the whole 3.3–4.0 population — everyone *not* affected by #3978 — and the three critical findings all follow from that blast radius. Either switch at `>= 4.1`, or land the `--destination .`/`--force` fix first so the `tools` invocation is provably equivalent to what those users get today (I verified it is, with those flags).

jkube-kit/jkube-kit-spring-boot/src/main/java/org/eclipse/jkube/springboot/SpringBootLayeredJar.java
```diff
@@ -81,26 +102,92 @@ public List<String> listLayers() {
...
+        kitLogger.info("Extracted Spring Boot layers using jarmode=%s", jarMode);
+        return;
+      } catch (IOException ioException) {
+        kitLogger.debug("Failed with jarmode=%s: %s", jarMode, ioException.getMessage());
+      }
+    }
+
+    // Fallback: try both jarmodes
+    IOException lastException = null;
+    for (String fallbackJarMode : new String[]{JARMODE_TOOLS, JARMODE_LAYERTOOLS}) {
```
manusa: **The fallback re-runs the mode that just failed, and then reports the wrong exception.**

The `catch` above logs at debug and falls through, so when `determineJarMode()` returned `tools` and it failed, this loop issues the identical command again — into a directory the failed attempt may have just written to, which (see the `getExtractArgs` comment) makes the retry *guaranteed* to fail. Gate the loop on `jarMode == null`.

Separately, `lastException` is overwritten unconditionally, so the cause attached to the `IllegalStateException` at line 132 is always the **layertools** attempt. A Boot 4.1 user whose `tools` extraction failed for a real reason gets `Process 'java -Djarmode=layertools -jar app.jar extract' exited with status 1` — verbatim the misleading message from #3978 — with the actual cause visible only at debug level. Please rethrow the primary failure, or aggregate with `addSuppressed`.

Nit while here: the success at line 126 logs nothing, unlike line 112 — that is the one case where knowing which mode won actually matters.

jkube-kit/jkube-kit-spring-boot/src/main/java/org/eclipse/jkube/springboot/SpringBootLayeredJar.java
```diff
@@ -81,26 +102,92 @@ public List<String> listLayers() {
...
+      String[] parts = version.split("[.-]");
+      if (parts.length < 2) {
+        return false;
+      }
+      int major = Integer.parseInt(parts[0]);
+      int minor = Integer.parseInt(parts[1]);
+
+      return major > 3 || (major == 3 && minor >= 3);
+    } catch (NumberFormatException e) {
+      kitLogger.debug("Unable to parse Spring Boot version: %s", version, e);
```
manusa: `KitLogger.debug` is `(String format, Object... params)`, and `AnsiUtil.format` only unwraps a `Throwable` when it is the **sole** param (`params.length == 1 && params[0] instanceof Throwable`). With two params this falls through to `String.format`, which ignores the surplus argument, so `e` is never logged.

Suggest `debug("Unable to parse Spring Boot version %s: %s", version, e.getMessage())`.

jkube-kit/jkube-kit-spring-boot/src/main/java/org/eclipse/jkube/springboot/SpringBootLayeredJar.java
```diff
@@ -65,6 +69,23 @@ public String getMainClass() {
     return null;
   }
 
+  public Optional<String> getSpringBootVersion() {
```
manusa, line 56: This is character-for-character `getMainClass()` with a different attribute name. Both parse `MANIFEST.MF` with `java.util.Properties`, which reads ISO-8859-1, treats backslash as an escape, and mangles the manifest's 72-byte continuation lines. Since you are adding a second caller, worth extracting one helper on the proper parser:

```java
private Optional<String> manifestAttribute(String name) {
  try (JarFile jarFile = new JarFile(layeredJar)) {
    return Optional.ofNullable(jarFile.getManifest())
        .map(m -> m.getMainAttributes().getValue(name));
  } catch (Exception e) {
    kitLogger.debug("Couldn't read %s from %s", name, layeredJar.getName());
  }
  return Optional.empty();
}
```

jkube-kit/jkube-kit-spring-boot/src/main/java/org/eclipse/jkube/springboot/generator/LayeredJarGenerator.java
```diff
@@ -70,4 +85,54 @@ public AssemblyConfiguration createAssemblyConfiguration(List<AssemblyFileSet> d
...
   }
+
+  /**
+   * Find the base directory containing layer subdirectories.
+   * tools jarmode extracts to <jar-basename>/<layer> structure,
+   * while layertools jarmode extracts directly to <layer> directories.
+   */
+  private File findLayerBaseDirectory(File buildPackageDirectory) {
+    // Check if layers exist directly in buildPackageDirectory (layertools behavior)
+    if (new File(buildPackageDirectory, DEPENDENCIES_LAYER).exists()) {
```
manusa: **A stale `dependencies/` from an earlier build silently wins over the layers just extracted.**

This root probe runs first, with no freshness check. I reproduced the upgrade path: a `target/` holding a previous layertools build (I dropped a `STALE_MARKER.txt` into `target/dependencies`), plus a new Spring Boot 4.1 jar. The extraction succeeds and correctly writes `target/app-1.0.0/{dependencies,...}` — but running this method verbatim against that tree returns **`target/`**, the stale directories, and the marker from the old build is inside the selected tree. The fresh output is ignored entirely.

Result: the image is assembled from the previous build's jars and classes — no error, no warning, exit code 0. Anyone who ran JKube once before upgrading Spring Boot hits this.

Passing `--destination .` removes the guessing entirely.

jkube-kit/jkube-kit-spring-boot/src/main/java/org/eclipse/jkube/springboot/generator/LayeredJarGenerator.java
```diff
@@ -70,4 +85,54 @@ public AssemblyConfiguration createAssemblyConfiguration(List<AssemblyFileSet> d
...
+      for (File subdir : subdirs) {
+        if (new File(subdir, DEPENDENCIES_LAYER).exists()) {
+          getLogger().debug("Found layers in subdirectory: %s", subdir.getName());
+          return subdir;
+        }
+      }
+    }
+
+    // Default to buildPackageDirectory if no layers found
+    return buildPackageDirectory;
```
manusa: **Hardcoding the `dependencies` layer name and then silently defaulting produces an image with no application content.**

All three probes key off the literal `DEPENDENCIES_LAYER`, and this line silently falls back. Reproduced with a custom `layers.xml` (supported via `spring-boot-maven-plugin`'s `<layers><configuration>`) whose layers are `libs`/`loader`/`snapshots`/`app`: extraction succeeds, and **no `dependencies` directory exists anywhere** under `target/`. Running this method verbatim then returns `buildPackageDirectory`, and the filesets built below resolve to:

```
target/libs       exists=false
target/loader     exists=false
target/snapshots  exists=false
target/app        exists=false
```

`AssemblyFileSetUtils.processAssemblyFileSet` returns `Collections.emptyList()` for a missing source directory, so the build succeeds and ships an empty `/deployments` — the symptom reported earlier in this PR.

If the heuristic survives, probe with `springBootLayeredJar.listLayers().get(0)` — already fetched two lines later and the authoritative source of layer names — and throw rather than silently defaulting.

jkube-kit/jkube-kit-spring-boot/src/main/java/org/eclipse/jkube/springboot/generator/LayeredJarGenerator.java
```diff
@@ -70,4 +85,54 @@ public AssemblyConfiguration createAssemblyConfiguration(List<AssemblyFileSet> d
...
+    if (jarBaseName != null) {
+      File expectedSubdir = new File(buildPackageDirectory, jarBaseName);
+      if (expectedSubdir.isDirectory() && new File(expectedSubdir, DEPENDENCIES_LAYER).exists()) {
+        getLogger().debug("Found layers in artifact-specific subdirectory: %s", jarBaseName);
+        return expectedSubdir;
+      }
+    }
+
+    // Fallback: search all subdirectories for layers
+    File[] subdirs = buildPackageDirectory.listFiles(File::isDirectory);
```
manusa: `File.listFiles` returns entries in an order `java.io.File#list` explicitly does not specify, so with more than one candidate the chosen layer base is filesystem-dependent. `File::isDirectory` also dereferences symlinks, and `getRelativePath` uses `getAbsolutePath()` rather than the canonical path, so a symlink under `target/` pointing outside the workspace would be accepted and its contents copied into the image. `target/classes/dependencies` (from a `src/main/resources/dependencies/` folder) is a plausible accidental match too.

Not a vulnerability — anything that can plant a directory under `target/` already runs inside the build — but a canonical-path containment check would be worth it if the scan stays.

jkube-kit/jkube-kit-spring-boot/src/main/java/org/eclipse/jkube/springboot/generator/LayeredJarGenerator.java
```diff
@@ -49,15 +53,26 @@ public AssemblyConfiguration createAssemblyConfiguration(List<AssemblyFileSet> d
     getLogger().info("Spring Boot layered jar detected");
     final List<Assembly> layerAssemblies = new ArrayList<>();
     layerAssemblies.add(Assembly.builder().id("jkube-includes").fileSets(defaultFileSets).build());
-    springBootLayeredJar.extractLayers(getProject().getBuildPackageDirectory());
 
+    File buildPackageDirectory = getProject().getBuildPackageDirectory();
+    getLogger().info("Build package directory: %s", buildPackageDirectory.getAbsolutePath());
```
manusa: Two absolute paths logged at INFO on every build. The pre-PR method emitted exactly one INFO line ("Spring Boot layered jar detected"), and lines 106/116 of this same new method log the same class of information at `debug`. Please demote both to `debug`.

jkube-kit/jkube-kit-spring-boot/src/test/java/org/eclipse/jkube/springboot/generator/LayeredJarGeneratorTest.java
```diff
@@ -0,0 +1,519 @@
...
+      // Then - Should successfully find layers after iterating through subdirs
+      assertThat(config.getLayers())
+          .hasSizeGreaterThan(1)
+          .extracting(Assembly::getId)
+          .contains("dependencies", "application");
+    }
+
+    @Test
+    @DisplayName("should prioritize jar-basename subdirectory (Spring Boot 3.3+ tools jarmode)")
+    void shouldPrioritizeJarBasenameSubdirectory() throws IOException {
```
manusa: **This suite cannot detect the feature it is testing being deleted.**

Every assertion in `FindLayerBaseDirectory` is on `Assembly::getId`, and IDs come only from `listLayers()` (generator lines 68/72). `findLayerBaseDirectory`'s return value influences only `AssemblyFileSet.directory`, which nothing in this PR asserts. Replace the method with `return buildPackageDirectory;` and all 9 tests here, the 5 in `CreateAssemblyConfiguration`, and the 3 in `LayeredJarGeneratorSubdirectoryTest` still pass.

This test is the sharpest case: it deliberately creates both `my-app-1.0.0/` and `other-dir/` "to test prioritization", but `contains("dependencies","application")` is identical for either outcome. The missing assertion:

```java
assertThat(config.getLayers()).filteredOn(a -> a.getId().equals("dependencies"))
  .flatExtracting(Assembly::getFileSets)
  .extracting(AssemblyFileSet::getDirectory)
  .containsExactly(new File("target/my-app-1.0.0/dependencies"));
```

That is the style `SpringBootGeneratorIntegrationTest` already uses for the layertools layout.

Related: the fixture is `Spring-Boot-Version: 2.7.14`, so `determineJarMode()` picks `layertools` and extraction lands at the root — `findLayerBaseDirectory` returns on its first branch in all 14 fixture-based tests. The tests named "...for tools jarmode (Spring Boot 3.3+)" never reach the tools path, `shouldIterateThroughMultipleSubdirsToFindDependencies` executes the loop zero times, and `whenNoDependenciesFound_...`, `whenSubdirsIsNull_...` and `whenOnlyFilesExist_...` have false premises. A `Spring-Boot-Version: 4.1.x` fixture with a `layers.idx` would fix the whole group.

jkube-kit/jkube-kit-spring-boot/src/test/java/org/eclipse/jkube/springboot/generator/LayeredJarGeneratorSubdirectoryTest.java
```diff
@@ -0,0 +1,197 @@
...
+      // When - createAssemblyConfiguration is called
+      // It will call extractLayers (mocked to no-op), then findLayerBaseDirectory
+      // findLayerBaseDirectory should NOT find dependencies in root,
+      // so it will loop through subdirectories (LINES 91-98) and find appSubdir
+      AssemblyConfiguration config = generator.createAssemblyConfiguration(Collections.emptyList());
+
+      // Then - Should successfully create assembly with layers from subdirectory
+      assertThat(config.getLayers())
+          .hasSizeGreaterThan(1) // jkube-includes + actual layers
+          .extracting(Assembly::getId)
```
manusa: Same gap as `LayeredJarGeneratorTest`: `listLayers()` is mocked to a fixed list, so asserting on `Assembly::getId` re-reads the stub and says nothing about which base directory was chosen. Assert `AssemblyFileSet::getDirectory` instead — that is the only observable this method affects.

Also, per the project's testing guidance ("minimize mocks; prefer real objects for internal classes"), `mockConstruction` on `SpringBootLayeredJar` is only needed because `LayeredJarGenerator` has no seam. Making `findLayerBaseDirectory` package-private and testing it directly against a fabricated tree removes the mock *and* the JVM fork.

jkube-kit/jkube-kit-spring-boot/src/test/java/org/eclipse/jkube/springboot/SpringBootLayeredJarFallbackTest.java
```diff
@@ -0,0 +1,149 @@
...
+   */
+  private static class TestableSpringBootLayeredJar extends SpringBootLayeredJar {
+    final List<String> attemptedJarModes = new ArrayList<>();
+
+    TestableSpringBootLayeredJar(File layeredJar, KitLogger kitLogger) {
+      super(layeredJar, kitLogger);
+    }
+
+    @Override
+    public void extractLayers(File extractionDir) {
```
manusa: **This override replaces the method the file says it verifies, so no assertion here observes production code.**

`Collections.addAll(attemptedJarModes, JARMODE_TOOLS, JARMODE_LAYERTOOLS)` at line 145 is the very data the assertion at line 73 reads back; reversing the production loop order, or deleting the loop entirely, leaves this file green. The only production code reached is `determineJarMode()`, which is already covered by `SpringBootVersionComparisonTest`.

The parameterized test at line 78 is also wrong about the contract: its name says "should use {1} **without fallback**", but production falls through on `IOException`, so for a 2.7.14 jar whose first attempt fails the real sequence is `layertools -> tools -> layertools`.

Suggestion: add a package-private seam on `SpringBootLayeredJar` (an injectable command factory, or a `List<String[]> plannedInvocations()`) and assert the recorded argv of the *real* `extractLayers`. Nothing in this PR currently asserts the emitted command line, which is exactly what #3978 is about.

jkube-kit/jkube-kit-spring-boot/src/test/java/org/eclipse/jkube/springboot/SpringBootLayeredJarTest.java
```diff
@@ -178,4 +181,271 @@ void extractLayers_whenJarInvalid_thenThrowException() {
...
+      try (JarOutputStream jarOutputStream = new JarOutputStream(Files.newOutputStream(jarFile.toPath()), manifest)) {
+        jarOutputStream.putNextEntry(new JarEntry("BOOT-INF/layers.idx"));
+      }
+      return jarFile;
+    }
+  }
+
+  @Nested
+  @DisplayName("extractLayers with version-specific jarmode")
+  class ExtractLayersWithJarMode {
```
manusa: This nested class is named "extractLayers with version-specific jarmode" but never calls `extractLayers`; its single assertion duplicates `GetSpringBootVersion.withValidVersion` above, differing only by adding `3.2.0`. As the body notes, the interesting behaviour is not covered — so a regression where `LayerToolsCommand` ignored its `jarMode` field would leave this green.

Suggest folding `3.2.0` into the `@ValueSource` above and dropping the class, or making it real by asserting the argv.

jkube-kit/jkube-kit-spring-boot/src/test/java/org/eclipse/jkube/springboot/SpringBootVersionComparisonTest.java
```diff
@@ -0,0 +1,157 @@
...
+    "3.3.0-M1, true",
+    "3.2.9-SNAPSHOT, false",
+    "4.1.1, true",
+    "5.0.0, true",
+    "10.0.0, true"
+  })
+  @DisplayName("version comparison")
+  void versionComparison(String version, boolean expected) throws Exception {
+    // Given
+    final File jarFile = createJarWithVersion(version);
```
manusa: `versionComparison` builds a jar per parameter (17 in total) that is never read — `isVersion330OrNewer(version)` takes the `String`. The two tests just below correctly use a non-existent `new File(projectDir, "test.jar")`. Suggest constructing once in `@BeforeEach` and dropping `throws Exception`.

Also: no case reaches the `catch (NumberFormatException)` branch — `"invalid"` and `"3"` both return via the `parts.length < 2` guard. Adding `"3.x, false"` would cover it.

manusa (changes requested):
Thanks for tackling this — the diagnosis is right, and I confirmed independently that `layertools` is genuinely gone in 4.1.

**How I reviewed this:** I did not build the PR (my tree is the base branch), so nothing below asserts a jkube test outcome. Instead I reproduced the Spring Boot side end to end: I assembled genuinely bootable layered fat jars (loader classes at the root, `spring-boot-jarmode-tools` STORED under `BOOT-INF/lib`, real `classpath.idx`/`layers.idx`) for Boot **3.5.0** and **4.1.1**, and ran `java -Djarmode=… -jar app.jar` through the real `JarLauncher` → `JarModeRunner` → `ToolsJarMode` path. I then executed this PR's `findLayerBaseDirectory`/`getJarBaseName` **verbatim** (only the `KitLogger` call stubbed — they touch nothing but `java.io.File`) against that real extraction output.

Version sweep, read from the Central artifacts' `META-INF/spring.factories`:

| `spring-boot-jarmode-tools` | 3.3.0 | 3.4.13 | 3.5.0 | 4.0.8 | 4.1.0 | 4.1.1 |
|---|---|---|---|---|---|---|
| `LayerToolsJarMode` registered | yes | yes | yes | yes | **no** | **no** |
| `--destination` / `--force` / `--launcher` / `--layers` | yes | yes | yes | yes | yes | yes |

Reproductions:

| Scenario (real fat jar, real launcher) | Result |
|---|---|
| Boot 4.1.1, `-Djarmode=layertools … extract` (what JKube 1.20 emits) | `Error: Unsupported jarmode 'layertools'`, exit 1 — the issue reproduced verbatim |
| Boot 4.1.1, **this PR's argv** `tools extract --launcher --layers`, build #1 | exit 0 → lands in `target/app-1.0.0/<layer>/` |
| Boot 4.1.1, **this PR's argv**, build #2 (no `mvn clean`) | **exit 1**, `JarModeErrorException: …/app-1.0.0 already exists and is not empty` |
| Boot 4.1.1, `… --destination . --force` ×3 | exit 0, 0, 0 → lands in `target/<layer>/` |
| Boot 3.5.0, `layertools extract` vs `tools … --destination . --force` | **identical file trees** (`diff` clean) |
| Boot 4.1.1, custom `layers.xml` (`libs/loader/snapshots/app`) | extracts fine; **zero** `dependencies` directories anywhere |

## Strengths

- The diagnosis is correct: the base code's hardcoded `-Djarmode=layertools` produces exactly the reported error on 4.1.
- **`--launcher --layers` is the right flag pair.** With `--launcher`, `ExtractCommand` takes the `extractArchive` branch and emits `<dest>/<layer>/<entry path>`; I verified the resulting tree is identical to legacy `layertools`. Loader classes are retained, so `JAVA_MAIN_CLASS=…JarLauncher` stays valid — this addresses the earlier "thin-jar layout / empty `/deployments`" report.
- Reading `Spring-Boot-Version` from the manifest is the right signal, and `determineJarMode()`/`isVersion330OrNewer()` are genuinely well covered, including milestone and snapshot qualifiers.
- Legacy projects are unaffected: for a layertools-shaped `target/`, `findLayerBaseDirectory` correctly returns the root (verified by executing it).
- Nothing security-relevant is introduced: the jarmode comes from two `static final` constants and is never interpolated from manifest data, `ExternalCommand` uses `Runtime.exec(String[])` (argv, no shell), and `getJarBaseName` cannot escape `buildPackageDirectory`.

## Issues

### Critical

**C1 — `getExtractArgs` omits `--destination .` and `--force`; the second build of any Boot >= 3.3 project fails.** (`SpringBootLayeredJar.java:141`)
Spring's own `layertools` compat shim (`ExtractLayersCommand`) delegates to `ExtractCommand` with exactly `--destination "." --layers --launcher --force`. This PR emits that command minus the last two. Reproduced above: build #1 succeeds into `target/app-1.0.0/`, build #2 fails with *"already exists and is not empty"*. JKube then retries `tools` (fails identically) and finally `layertools`, which on 4.1+ does not exist — three JVM forks, then a hard failure, on the very version this PR fixes.

**C2 — A stale `target/dependencies` silently wins over the freshly extracted layers.** (`LayeredJarGenerator.java:96`)
Reproduced: a `target/` holding a previous layertools build plus a new Boot 4.1 jar. Extraction succeeds into `target/app-1.0.0/`, but the resolution logic returns `target/` — the stale directories — and a marker file from the old build sits inside the selected tree. The image ships the previous build's content, with no error and exit code 0.

**C3 — The hardcoded `dependencies` probe plus a silent default produces an empty image.** (`LayeredJarGenerator.java:96,105,115,123`)
Reproduced with a custom `layers.xml`: extraction succeeds with layers `libs/loader/snapshots/app` and no `dependencies` directory anywhere, so all three probes miss and `:123` silently returns `buildPackageDirectory`. Every resulting fileset directory resolves to `exists=false`, and `AssemblyFileSetUtils.processAssemblyFileSet` returns `Collections.emptyList()` for a missing source dir — silently. Build succeeds, `/deployments` is empty.

**C4 — The new generator tests cannot detect the feature being removed.** (`LayeredJarGeneratorTest.java:245-450`, `LayeredJarGeneratorSubdirectoryTest.java:81-188`)
Every assertion in both `findLayerBaseDirectory` suites is on `Assembly::getId`, which comes only from `listLayers()`. `findLayerBaseDirectory` influences only `AssemblyFileSet.directory`, and nothing in the diff asserts it. Replacing the whole method with `return buildPackageDirectory;` leaves all 17 tests green.

### Important

- **I1** — the fallback re-runs the mode that just failed, and the cause attached to the thrown `IllegalStateException` is always the *last* (layertools) attempt, reproducing the misleading message from #3978. (`SpringBootLayeredJar.java:114-132`)
- **I2** — the 3.3.0 threshold is wider than the bug: `layertools` works through 4.0.8, only 4.1 removed it. As written, the command, layout and directory resolution change for everyone on 3.3–4.0, i.e. everyone *not* affected by #3978 — and C1/C2/C3 all follow from that blast radius. (`SpringBootLayeredJar.java:157`)
- **I3** — `SpringBootLayeredJarFallbackTest` overrides the method it claims to verify and asserts against its own stub; its "without fallback" contract is not what production does. (`SpringBootLayeredJarFallbackTest.java:137`)
- **I4** — there is no Spring Boot 3.3+ fixture in the module; the only layered jar is `Spring-Boot-Version: 2.7.14`, so every test named "…for tools jarmode (Spring Boot 3.3+)" actually runs `layertools` and returns on the first branch.
- **I5** — unit tests now fork roughly 21 real JVMs (up from 5); none of these tests is about subprocess execution.
- **I6** — the description and CHANGELOG cover about half the change: no mention of `--launcher --layers`, the `LayeredJarGenerator` rewrite, the new INFO logging, or the layout move to `target/<jar-basename>/<layer>`. "Fully backward compatible" is contradicted by C1 and that layout change.
- **I7** — `ExtractLayersWithJarMode` never calls `extractLayers`; it duplicates `GetSpringBootVersion`. (`SpringBootLayeredJarTest.java:250`)

### Minor

- **M1** `LayeredJarGenerator.java:112` — the blind subdirectory scan is order-dependent (`File.listFiles` order is unspecified), follows symlinks, and has no containment check.
- **M2** `LayeredJarGenerator.java:58,64` — two absolute paths logged at INFO on every build, inconsistent with `:106`/`:116` in the same method.
- **M3** `SpringBootLayeredJar.java:126` — the fallback success path logs nothing.
- **M4** `SpringBootLayeredJar.java:159` — the exception is passed as a surplus format arg and silently dropped.
- **M5** `SpringBootLayeredJar.java:72-87` — verbatim copy of `getMainClass()`; `JarFile.getManifest()` collapses both.
- **M6** `SpringBootLayeredJar.java:165-173` — `null` sentinel two methods after `Optional` was introduced for that concept.
- **M7** `SpringBootLayeredJar.java:137,176` — `LayerToolsCommand` now runs `tools` too; the comment says "4.x" but the code applies from 3.3; stray double blank line at `:174`.
- **M8** `SpringBootVersionComparisonTest.java:64` — builds 17 jars that are never read; the `NumberFormatException` branch is unreachable from any case.
- **M9** test hygiene: unused `java.lang.reflect.Method` import, missing trailing newlines, stale "LINES 91-98" comments, `createJarWithVersion` duplicated three times, dead `inSubdirectory` parameter, new top-level classes where the module convention is `@Nested`, `mockConstruction` on a first-party class.
- **M10** `SpringBootLayeredJarTest.java:297-320` — `getExtractArgs("unknown")`/`(null)` test states production cannot reach.

### Investigated and dropped

- **Executable-jar regression** (`<executable>true</executable>`): tested both modes on 3.5.0 — both fail identically, because the layertools shim delegates to the same `ExtractCommand.checkJarCompatibility`. No delta from this PR.
- **`--layers` with no value being malformed**: refuted, it is declared optional-value and means "all layers".
- **Gradle-specific breakage**: refuted, `build/libs` behaves the same as `target/`.
- **`isVersion330OrNewer` parsing fragility**: largely refuted; only the null-argument nit survives (M6).

## Assessment

The diagnosis is right and the flags are right, but as shipped the PR does not fix #3978 beyond the first build (C1), and adds two silent-wrong-output paths (C2, C3) for the 3.3–4.0 majority who were never affected by the bug.

The single highest-value change is **C1**: adding `--destination . --force` reproduces Spring's own shim, is idempotent across repeated builds, produces output identical to `layertools`, and restores today's on-disk layout — which lets `findLayerBaseDirectory`/`getJarBaseName` and C2/C3/M1 be deleted outright. Then **I2** and **I1**. The test suite needs one assertion on `AssemblyFileSet.directory` and one on the emitted argv (C4, I3, I4) before it can protect any of this.

jkube-kit/jkube-kit-spring-boot/src/main/java/org/eclipse/jkube/springboot/generator/LayeredJarGenerator.java
ash-thakur-rh: yes, agree. I was not aware about this flag. Since with this `--destination .` flag is there which directly extract layer in build directly. I think it is better to use static dir inside of build dir to extract the layers in. This way we will be not polluting the build directly with multiple layer folders, as in future user can utilize similar folder names.

jkube-kit/jkube-kit-spring-boot/src/main/java/org/eclipse/jkube/springboot/generator/LayeredJarGenerator.java
ash-thakur-rh: I think this is not required as we control the layer extraction destination path. I will remove the layer finding and related methods.

jkube-kit/jkube-kit-spring-boot/src/main/java/org/eclipse/jkube/springboot/generator/LayeredJarGenerator.java
ash-thakur-rh: done

jkube-kit/jkube-kit-spring-boot/src/test/java/org/eclipse/jkube/springboot/generator/LayeredJarGeneratorSubdirectoryTest.java
ash-thakur-rh: removing test as not required after destination option.

jkube-kit/jkube-kit-spring-boot/src/test/java/org/eclipse/jkube/springboot/SpringBootLayeredJarTest.java
ash-thakur-rh: done

jkube-kit/jkube-kit-spring-boot/src/test/java/org/eclipse/jkube/springboot/generator/LayeredJarGeneratorTest.java
ash-thakur-rh: These tests are now obsolete, need to test destination now.

jkube-kit/jkube-kit-spring-boot/src/main/java/org/eclipse/jkube/springboot/SpringBootLayeredJar.java
ash-thakur-rh: done, now using both destination and new command with sb >= 4.1

jkube-kit/jkube-kit-spring-boot/src/main/java/org/eclipse/jkube/springboot/SpringBootLayeredJar.java
ash-thakur-rh: done

jkube-kit/jkube-kit-spring-boot/src/main/java/org/eclipse/jkube/springboot/SpringBootLayeredJar.java
ash-thakur-rh: done

jkube-kit/jkube-kit-spring-boot/src/main/java/org/eclipse/jkube/springboot/SpringBootLayeredJar.java
ash-thakur-rh: done

jkube-kit/jkube-kit-spring-boot/src/main/java/org/eclipse/jkube/springboot/SpringBootLayeredJar.java
ash-thakur-rh: done

jkube-kit/jkube-kit-spring-boot/src/test/java/org/eclipse/jkube/springboot/SpringBootLayeredJarFallbackTest.java
ash-thakur-rh: done

manusa (approved):
LGTM, thx!

